PR #3, opened 2022-06-28
build: add GH Action definition

vburckhardt:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  call-terraform-ci-pipeline:
+    uses: terraform-ibm-modules/common-dev-assets/.github/workflows/terraform-module-ci.yml@main
```
toddgiguere, line 16: Reusable workflows are no longer in common-dev-assets, we have moved them to common-pipeline-assets and pin them to a specific version. You can use this:
`terraform-ibm-modules/common-pipeline-assets/.github/workflows/common-terraform-module-ci.yml@v1.2.0`

.github/workflows/release.yml
```diff
@@ -0,0 +1,16 @@
...
+    branches: [main]
+    types:
+      - completed
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  call-terraform-release-pipeline:
+    uses: terraform-ibm-modules/common-dev-assets/.github/workflows/release.yml@main
```
toddgiguere, line 15: The release needs to point to common-pipeline-assets (see above), and also will require a conditional statement so that release will only trigger if CI was successful. You can copy the file we have in the template: https://github.com/terraform-ibm-modules/terraform-ibm-module-template/blob/main/.github/workflows/release.yml

toddgiguere (changes requested):

.github/workflows/ci.yml
vburckhardt: @toddgiguere - Ok this is coming from the module template at https://github.com/terraform-ibm-modules/terraform-ibm-module-template, so we should update there as well.

vburckhardt: My mistake actually - I took an older version of the template.

PR #48, opened 2022-08-17
test: introduce override example, and corresponding test

vburckhardt:
### Description

Basic example showing how to use override.json, used to introduce a test leveraging this capability mostly

**Check the relevant boxes:**
- [ ] Bug fix (nonbreaking change that fixes an issue)
- [ ] New feature (nonbreaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Examples or tests (addition or updates of examples or tests)
- [ ] Documentation update
- [ ] CI related update (pipeline, etc.)

### Checklist

- [ ] If relevant, a test for the change has been added or updated as part of this PR.
- [ ] If relevant, documentation for the change has been added or updated as part of this PR.

### Merge

- Merge using "Squash and merge".
- Make sure to use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents. The commit message determines whether a new version of the modules needs to be released, and if so, which semver number to use).


ocofaigh (changes requested):
@vburckhardt you need to bump your version of common-dev-assets for pipeline to run. You have included an old version in the PR

vburckhardt:
Correct, not ready to review yet. Planning a few more iteration. I'll submit review request when ready.

tests/pr_test.go
```diff
@@ -78,3 +80,45 @@ func TestRunUpgradeBasicExample(t *testing.T) {
...
+	t.Parallel()
+
+	options := setupOptionsOverride(t, "slz-json")
+
+	output, err := options.RunTestConsistency()
+	assert.Nil(t, err, "This should not have errored")
+	assert.NotNil(t, output, "Expected some output")
+}
+
+// Re-introduce when initial version is in master
```
ocofaigh, line 112: The source code of the module is already in master, so you could enable this now. however because there are static names used in the override.json, I am wondering if the two tests would start to clash if you did enable this? Also if two PRs were opened around the same time, they would also clash as well. Not sure what best way to handle it is, unless we some how dynamically generated the override json for every test run?

vburckhardt: The example that is added in this PR is not in main yet. Once this is merged, the upgrade test for this example can be enabled. The prefix should be used so there should not be clashes.

ocofaigh: ah yes right, the example needs to be in main too duh. 
Is prefix still used when using override.json though? For example, I see prefix defined in the override.json in places. Also won't things like these clash?
```
"resource_group": "slz-service-rg"
"collector_bucket_name": "atracker-bucket"
```

vburckhardt: @ocofaigh - The prefix is added to the values in the json file. For instance, when running the tests we end up with "slz-json-o5duxj-slz-service-rg" based on the logic at https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/add-custom-example/tests/pr_test.go#L105

examples/override-example/variables.tf
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+variable "region" {
+  description = "Region where VPC will be created. To find your VPC region, use `ibmcloud is regions` command to find available regions."
+  type        = string
+  default     = "us-south"
+}
+
+variable "resource_group" {
```
argeiger, line 28: Don't think this variable is used.

argeiger (changes requested):

examples/override-example/variables.tf
vburckhardt: Correct - coming from vsi quick start example. Fixed in both. We'll need to move to full checks eventually - currently using basics only https://github.com/terraform-ibm-modules/common-dev-assets/tree/main/module-assets/basic-pre-commit which do skip unused variable checks.

argeiger (dismissed):

vburckhardt:
Dependency on a fix in terratest wrapper https://github.com/terraform-ibm-modules/ibmcloud-terratest-wrapper/pull/91

vburckhardt:
@argeiger - Another review for latest change since last review.

PR #66, opened 2022-08-30
docs: update description for public ssh key

JonWoodlief:


patterns/vsi/variables.tf
```diff
@@ -30,7 +30,7 @@ variable "region" {
 }
 
 variable "ssh_public_key" {
-  description = "Public SSH Key for VSI creation. Must be a valid SSH key that does not already exist in the deployment region."
+  description = "Public SSH Key for VSI creation. Must be a valid SSH key that does not already exist in the deployment region. The Private SSH Key associated with the provided Public SSH Key may be needed by users or tooling requiring access to servers on your VPC."
```
vburckhardt, line 33: My take - mix for style and accuracy:
"The private SSH key associated with the provided public SSH key is used to establish a ssh session to VSIs on your VPC."

@SirSpidey - would value your input here as well

JonWoodlief: I think that sounds good.

The one thing I would want to add on top of it, is some sort of indication that the private key may be asked for later. I don't want users to pass their personal key, and then when they try to do an add-on (such as power) later, they're surprised they need to pass their personal private key

JonWoodlief: I want to steer users to pass a public key associated w/ a private key they're comfortable sharing

patterns/vsi/variables.tf
```diff
@@ -30,7 +30,7 @@ variable "region" {
...
   type        = string
   validation {
     error_message = "Public SSH Key must be a valid ssh rsa public key."
```
SirSpidey, line 36:
```suggestion
    error_message = "Make sure that the key is a valid SSH RSA public key."
```

SirSpidey (changes requested):
Saw one other suggested change in this PR: to the error message

patterns/vsi/variables.tf
ninahawley-ibm: @SirSpidey I like it, but if my understanding is correct we're supposed to stray away from using the VSI acronym and should spell it out instead. Also, I thought the SSH key can't exist in the specified region already - as in, it does not already exist in the specified region? Is that the case, or does the key need to be created outside of the specified region? 

My further tweak:

>A public SSH key to create and connect to virtual server instances. Use an SSH key that does not already exist in the specified region. The private SSH key for this public key is used to connect to virtual server instances on your VPC.

SirSpidey: 👍 @ninahawley-ibm keeping me honest about abbreviations. But I need someone else to weigh in on the region question. @JonWoodlief @vburckhardt ?

ocofaigh:
@JonWoodlief You may have to recreate this PR from a direct branch. We are currently working on a POC which will allow contribution from forks, but until then I have added you to the org as a collaborator, so please accept that invite, and create a new PR using direct branch. Thanks

PR #253, opened 2023-01-11
[WIP] feat: Integrate ocp-base with SLZ

Aashiq-J:
### Description

Please include a summary of the change and which issue is fixed. Also, please include the motivation and context for the change. List any dependencies that are required for this change.

Issue : https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/324

**Check the relevant boxes:**
- [ ] Bug fix (nonbreaking change that fixes an issue)
- [x] New feature (nonbreaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] CI related update (pipeline, etc.)

### Checklist

- [ ] If relevant, a test for the change has been added or updated as part of this PR.
- [x] If relevant, documentation for the change has been added or updated as part of this PR.

### Merge

- Merge using "Squash and merge".
- Make sure to use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents. The commit message determines whether a new version of the modules needs to be released, and if so, which semver number to use).

Release notes content

https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/extend-output/.docs/update-version.md


cluster.tf
```diff
@@ -1,104 +1,124 @@
...
-    content {
-      subnet_id = zones.value["id"]
-      name      = zones.value["zone"]
-    }
+module "workload_cluster" {
+  depends_on = [
+    module.vpc, module.observability_instances
+  ]
+  count             = length(module.dynamic_values.clusters_map) >= 1 ? 1 : 0
+  source            = "./ocp-all-inclusive"
```
vburckhardt, line 19: I think we should move this module in its own repo, just like what we do internally - so that we can have its own lifecycle, CI, etc independently from the rest of the SLZ module.

ocp-all-inclusive/main.tf
```diff
@@ -0,0 +1,117 @@
...
+  # If encryption enabled generate kms config to be passed to cluster
+  kms_config = var.existing_key_protect_instance_guid != null && var.existing_key_protect_root_key_id != null ? {
+    crk_id           = var.existing_key_protect_root_key_id
+    instance_id      = var.existing_key_protect_instance_guid
+    private_endpoint = var.key_protect_use_private_endpoint
+  } : null
+}
+
+module "ocp_base" {
+  source                          = "git::https://github.ibm.com/GoldenEye/base-ocp-vpc-module.git?ref=4.0.1"
```
vburckhardt, line 23: base-ocp-vpc-module should be pushed out to terraform-ibm-module and referenced from there

ocp-all-inclusive/main.tf
```diff
@@ -0,0 +1,117 @@
...
+  logdna_sts_validate_condition = var.logdna_sts_instance_name != null && var.logdna_sts_ingestion_key == null
+  logdna_sts_validate_msg       = "A value for var.logdna_sts_ingestion_key must be passed when providing a value for var.logdna_sts_instance_name"
+  # tflint-ignore: terraform_unused_declarations
+  logdna_sts_validate_check = regex("^${local.logdna_sts_validate_msg}$", (!local.logdna_sts_validate_condition ? local.logdna_sts_validate_msg : ""))
+}
+
+module "observability_agents" {
+  # cluster-proxy required so observability images can be pulled from public registry
+  count                        = local.run_observability_agents_module == true ? 1 : 0
+  source                       = "git::https://github.ibm.com/GoldenEye/observability-agents-module?ref=2.6.0"
```
vburckhardt, line 78: Same as above, this module should be pushed out in its own repo in terraform-ibm-module

vburckhardt (commented):
Hi @Aashiq-J - thanks for the PR.

I think the approach should be as follows:
- Create new repos under terraform-ibm-module for the ocp-all-inclusive, ocp-base, ocp-observability-agent modules. This is the same repo structure as we currently have internally. The idea here is to enable the use of those modules on top of SLZ VPCs, but also usable on top of any VPC, including the VPC structure used internally
- In the ocp-all-inclusive example folder, have an example of deploying the ocp-all-inclusive module on top of an existing SLZ VPC. See the VSI example here https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vsi/blob/main/examples/default/main.tf as a source of inspiration. You could use https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/tree/main/examples/no-compute-example as a base to create the VPC.

cluster.tf
```diff
@@ -1,104 +1,92 @@
...
-  pod_subnet         = each.value.pod_subnet
-  service_subnet     = each.value.service_subnet
 
-  dynamic "zones" {
-    for_each = each.value.subnets
-    content {
-      subnet_id = zones.value["id"]
-      name      = zones.value["zone"]
-    }
+module "workload_cluster" {
```
argeiger, line 14: How would the user update the cluster (either patch version or major)?

cluster.tf
```diff
@@ -1,104 +1,92 @@
...
+  management_cluster = length(module.dynamic_values.clusters_map) == 2 ? module.dynamic_values.clusters_map["${var.prefix}-management-cluster"] : null
+}
 
-  dynamic "zones" {
-    for_each = each.value.subnets
-    content {
-      subnet_id = zones.value["id"]
-      name      = zones.value["zone"]
-    }
+module "management_cluster" {
```
argeiger, line 56: Instead of creating a new module, should you do a for_each on the cluster_map?  What if the user wants a cluster in the management vpc vs workload vpc?

argeiger (changes requested):

cluster.tf
Aashiq-J: Initially, I tried using for_each on the cluster_map, but that creates a list of clusters as output and that list cannot be passed to the helm provider to configure the cluster.
a helm provider can be used only to configure a single cluster, looping the helm provider to configure the two clusters is not possible. We are using the helm provider to install the observability agents.

But, I have updated the logic to have a management cluster created without a workload cluster.

cluster.tf
Aashiq-J: Updated the code to fetch the ocp_version from the override.json

cluster.tf
argeiger: For the observability, can we use this resource - https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/ob_logging

argeiger (changes requested):

cluster.tf
```diff
@@ -1,104 +1,92 @@
...
-  for_each          = local.worker_pools_map
-  vpc_id            = each.value.vpc_id
-  resource_group_id = local.resource_groups[each.value.resource_group]
-  entitlement       = each.value.entitlement
-  cluster           = ibm_container_vpc_cluster.cluster[each.value.cluster_name].id
-  worker_pool_name  = each.value.name
-  flavor            = each.value.flavor
-  worker_count      = each.value.workers_per_subnet
+locals {
+  management_cluster = lookup(module.dynamic_values.clusters_map, "${var.prefix}-management-cluster", false) != false ? module.dynamic_values.clusters_map["${var.prefix}-management-cluster"] : null
```
argeiger, line 53: This will break if the user changes the name

argeiger (changes requested):

cluster.tf
Aashiq-J: The observability instances are created in the observability.tf, but the logdna and the sysdig agents are installed on the cluster using the helm_release provider for which we had to separate out the creation of each cluster.

observability.tf
```diff
@@ -0,0 +1,24 @@
+locals {
+  observability_map = module.dynamic_values.clusters_map
+}
+
+
+##############################################################################
+# Observability Instances (LogDNA + Sysdig)
+##############################################################################
+
+module "observability_instances" {
```
vburckhardt, line 10: I would just leave the observability stuff out for now. It's a large topic on its own, and would require further discussions. Let's just focus this PR on consolidating the base-ocp logic.

cluster.tf
```diff
@@ -1,104 +1,89 @@
...
-      crk_id           = module.key_management.key_map[kms_config.value.crk_name].key_id
-      instance_id      = module.key_management.key_management_guid
-      private_endpoint = kms_config.value.private_endpoint
-    }
+module "cluster_1" {
+  depends_on = [
+    module.vpc, module.observability_instances
+  ]
+  count             = local.cluster_1 != null ? 1 : 0
+  source            = "git::https://github.com/terraform-ibm-modules/terraform-ibm-ocp-all-inclusive.git?ref=v1.0.0"
```
vburckhardt, line 20: Just use the ocp-base module (https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc) for now to simplify - so that we can narrow down the scope of this PR to just consolidating the ocp logic without adding new capabilities. We can add the observability layer in a subsequent step.

cluster.tf
```diff
@@ -1,104 +1,89 @@
...
-  }
 
-  dynamic "kms_config" {
-    for_each = each.value.kms_config == null ? [] : [each.value.kms_config]
-    content {
-      crk_id           = module.key_management.key_map[kms_config.value.crk_name].key_id
-      instance_id      = module.key_management.key_management_guid
-      private_endpoint = kms_config.value.private_endpoint
-    }
+module "cluster_1" {
```
vburckhardt, line 15: I think we're likely losing some of the capabilities from the original here - could we loop on the for_each = local.clusters_map like in the original and call the ocp_base module for_each times - mapping the values from the local.clusters_map to the input of the ocp-base module?

vburckhardt (changes requested):
Hi Aashiq, I put some comments in the PR. I think generally here the focus would be to just replace the existing ocp logic with a call to the module https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc, without adding more capabilities (such as observability stuff). Hope this makes things a bit simpler for you.

cluster.tf
Aashiq-J: Yea, its possible to use for_each with the cluster_map and have the base-ocp to create the cluster.

variables.tf
```diff
@@ -751,36 +757,38 @@ variable "clusters" {
...
-      machine_type       = string           # Worker node flavor
-      kube_type          = string           # iks or openshift
-      kube_version       = optional(string) # Can be a version from `ibmcloud ks versions` or `latest`
-      entitlement        = optional(string) # entitlement option for openshift
-      pod_subnet         = optional(string) # Portable subnet for pods
-      service_subnet     = optional(string) # Portable subnet for services
-      resource_group     = string           # Resource Group used for cluster
-      cos_name           = optional(string) # Name of COS instance Required only for OpenShift clusters
-      update_all_workers = optional(bool)   # If true force workers to update
+      name                    = string           # Name of Cluster
```
argeiger, line 760: Changes to the variables names will break the SLZ GUI.  The new optional ones are fine but the existing ones being changed will break it

@vburckhardt

argeiger (changes requested):

variables.tf
Aashiq-J: @argeiger , @vburckhardt 

I have reverted the changes to the variable names in the `worker_pools` list within the `clusters` variable. Regarding the variable `ocp_version` should I revert it to `kube_version`? Because we are not deploying k8s cluster anymore using the current code, so I guess `ocp_version` is the more suitable variable name.

Aashiq-J: I have added the `kube_version` back, but it doesn't serve any purpose. The `ocp_version` variable is being used to determine the ocp version.

cluster.tf
```diff
@@ -1,104 +1,53 @@
...
-resource "ibm_container_vpc_cluster" "cluster" {
+
+module "cluster" {
+  depends_on = [
+    module.vpc
+  ]
   for_each          = local.clusters_map
-  name              = "${var.prefix}-${each.value.name}"
-  vpc_id            = each.value.vpc_id
+  source            = "git::https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc.git?ref=v2.1.1"
```
argeiger, line 19: I tested with the following configuration - 
```
"clusters": [
    {
      "cos_name": "cos",
      "entitlement": null,
      "kube_type": "openshift",
      "kube_version": "default",
      "machine_type": "bx2.16x64",
      "name": "workload-cluster",
      "resource_group": "workload-rg",
      "kms_config": {
        "private_endpoint": true,
        "crk_name": "slz-roks-key"
      },
      "subnet_names": [
        "vsi-zone-1",
        "vsi-zone-2"
      ],
      "update_all_workers": false,
      "vpc_name": "workload",
      "worker_pools": [
        {
          "entitlement": "cloud_pak",
          "flavor": "bx2.16x64",
          "name": "default",
          "subnet_names": [
            "vsi-zone-1",
            "vsi-zone-2"
          ],
          "vpc_name": "workload",
          "workers_per_subnet": 1
        }
      ],
      "workers_per_subnet": 1
    }
  ]
```

I would expect the "default" (ibm_container_vpc_cluster) be from the high level information and the additional worker pool from the worker_pools information.  In this case, I expect 2 pools from this configuration

cluster.tf
```diff
@@ -1,104 +1,53 @@
...
-  flavor            = each.value.flavor
-  worker_count      = each.value.workers_per_subnet
-
-  dynamic "zones" {
-    for_each = each.value.subnets
-    content {
-      subnet_id = zones.value["id"]
-      name      = zones.value["zone"]
+  vpc_subnets = {
+    vsi-zone-1 = [
```
argeiger, line 26: What if they have multiple worker pools using different subnets?  Don't you need a map per pool?

cluster.tf
```diff
@@ -1,104 +1,53 @@
...
+    for pool in each.value.worker_pools :
+    {
+      subnet_prefix    = pool.subnet_names[0]
+      pool_name        = pool.name
+      machine_type     = pool.flavor
+      workers_per_zone = pool.workers_per_subnet
     }
+  ]
+  ocp_version             = each.value.ocp_version
+  tags                    = var.resource_tags
```
argeiger, line 45: Missing entitlement so looks to default to cloud_paks which most customers don't have

argeiger (changes requested):

cluster.tf
```diff
@@ -1,104 +1,53 @@
...
+  use_existing_cos        = true
+  disable_public_endpoint = each.value.disable_public_endpoint
+  existing_cos_id         = each.value.cos_instance_crn
+  kms_config = {
```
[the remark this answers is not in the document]
Aashiq-J: Support is being added to the base-ocp-module . https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc/pull/76

cluster.tf
Aashiq-J: Hi @argeiger,

@vburckhardt had suggested to have the single worker pool which has the nodes spread across different subnets/zone for high availability.
To support the multiple worker pools using different subnets, the base-ocp code has to be completely revamped.

cluster.tf
Aashiq-J: added this parameter.

cluster.tf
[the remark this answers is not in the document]
Aashiq-J: added support for using key protect key for boot volume.

cluster.tf
```diff
@@ -1,104 +1,71 @@
...
-resource "ibm_container_vpc_cluster" "cluster" {
+
+module "cluster" {
+  depends_on = [
+    module.vpc
+  ]
   for_each          = local.clusters_map
-  name              = "${var.prefix}-${each.value.name}"
-  vpc_id            = each.value.vpc_id
+  source            = "git::https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc.git?ref=v2.3.0"
```
argeiger, line 19: Need to have disable_public_service_endpoint set to "true" by default and ability to pass in

argeiger (changes requested):

patterns/mixed/variables.tf
```diff
@@ -152,8 +152,8 @@ variable "cluster_zones" {
   }
 }
 
-variable "kube_version" {
-  description = "Kubernetes version to use for cluster. To get available versions, use the IBM Cloud CLI command `ibmcloud ks versions`. To use the latest version, leave as latest. Updates to the latest versions may force this to change."
+variable "ocp_version" {
+  description = "Kubernetes version to use for cluster. To get available versions, use the IBM Cloud CLI command `ibmcloud ks versions`. To use the default version, leave as default. Updates to the default versions may force this to change."
```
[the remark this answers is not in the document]
argeiger: We don't want to break the json/gui if possible

cluster.tf
```diff
@@ -1,104 +1,72 @@
...
+        boot_volume_encryption_kms_config = {
+          crk             = module.key_management.key_map[each.value.kms_config.crk_name].key_id
+          kms_instance_id = module.key_management.key_management_guid
+        }
+      }
+  ])
+  ocp_version                     = each.value.kube_version
+  tags                            = var.resource_tags
+  use_existing_cos                = true
+  disable_public_endpoint         = each.value.disable_public_endpoint
```
argeiger, line 65: You need to look this up and then either set it if it has a value or default it to "true"

argeiger (changes requested):

cluster.tf
```diff
@@ -1,104 +1,72 @@
...
+  verify_worker_network_readiness = !(each.value.disable_public_endpoint)
```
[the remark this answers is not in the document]
Aashiq-J: If both `disable_public_endpoint ` and  `verify_worker_network_readiness ` are set to true, the terraform apply will fail. To verify_worker_network_readiness the code will try to run a data block to get the kube-config and if the public endpoint is disabled the terraform code will fail execution.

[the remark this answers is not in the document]
Aashiq-J: The default value for `verify_worker_network_readiness ` has to false, If the runtime does not have access to the kube cluster to run kubectl commands, this should be set to false. when the pr pipeline runs it wouldn't have access to the private kube-cluster and the pr pipeline will fail.

I can add the option to take the value from override.json.

variables.tf
```diff
@@ -843,6 +855,30 @@ variable "wait_till" {
...
+      var.kube_version == "4.11",
+      var.kube_version == "4.12"
+    ])
+    error_message = "The specified kube_version is not one of the validated versions."
+  }
+}
+
+variable "resource_tags" {
+  type        = list(string)
+  description = "Optional list of tags to be added to created resources"
```
ocofaigh, line 878: This implies all resources will be tagged with these tags. But as far as I can see its only used by cluster?
So do we rename the variable here? Also what about resource tagging for other resources deployed?

variables.tf
```diff
@@ -843,6 +855,30 @@ variable "wait_till" {
...
 
+variable "kube_version" {
+  description = "The version of the OpenShift cluster that should be provisioned (format 4.x). This is only used during initial cluster provisioning, but ignored for future updates. If no value is passed, or the string 'latest' is passed, the current latest OCP version will be used."
+  type        = string
+  default     = null
+  validation {
+    condition = anytrue([
+      var.kube_version == null,
+      var.kube_version == "latest",
+      var.kube_version == "4.8",
```
ocofaigh, line 866: 4.8 is not supported by terraform-ibm-base-ocp-vpc module

variables.tf
```diff
@@ -843,6 +855,30 @@ variable "wait_till" {
   }
 }
 
+variable "kube_version" {
+  description = "The version of the OpenShift cluster that should be provisioned (format 4.x). This is only used during initial cluster provisioning, but ignored for future updates. If no value is passed, or the string 'latest' is passed, the current latest OCP version will be used."
+  type        = string
+  default     = null
+  validation {
```
ocofaigh, line 862: Don't we need to continue to support classic IKS? In which case, I think that OCP versions need to be passed like this: `"4.x_openshift"`
And classic IKS will be passed like `1.18`

We should discuss

ocofaigh:
@Aashiq-J Why was upgrade test skipped? I think we need to be able to support the upgrade case to ensure current consumers clusters do not get destroyed. I imagine we will need to use "moved" blocks for the cluster resources changing here

variables.tf
```diff
@@ -2,6 +2,12 @@
 # Account Variables
 ##############################################################################
 
+variable "ibmcloud_api_key" {
+  description = "The IBM Cloud platform API key needed to deploy IAM enabled resources."
```
ocofaigh, line 6: Actually this variable is only used by `terraform-ibm-base-ocp-vpc` and its used to create the containers apikey (if one does not exist already) so perhaps update the description here

variables.tf
Aashiq-J: Duplicate variable, removing this.

Aashiq-J:
Hi @ocofaigh , Initially there was many changes including many required variables and variable name changes, so had to skip upgrade test. But, later most of the changes were pushed to later phases and changed variable names were reverted.

Is there a way to run the upgrade test now? I tried creating a different branch and pull request but it took the all the commit messages from this branch which caused it to skip the upgrade test in the new pull request as well.

ocofaigh:
@Aashiq-J I actually created an internal issue to allow the ability to re-enable the upgrade test, but its not done yet. I haven't tried this, but maybe you can try squashing all the commits into one commit (that does not have skip upgrade test in it). something like this (not tested): `git reset $(git merge-base main $(git rev-parse --abbrev-ref HEAD))`

Aashiq-J:
@ocofaigh 
I ran the upgrade test and it will destroy the cluster and recreate it.
Screenshot 2023-04-03 at 11 33 40 AM

Aashiq-J:
Hi @ocofaigh ,
I also tried to use the `moved` block, but `moved` block can only be used when we have a `A single static variable reference is required`. If we dynamically try to create the cluster name for the moved block it gives the following error.
```
│ Error: Invalid expression
│ 
│   on ../../moved.tf line 7, in moved:
│    7:   from = ibm_container_vpc_cluster.cluster["${var.prefix}-management-cluster"]
│ 
│ A single static variable reference is required: only attribute access and indexing with constant keys. No calculations, function calls, template expressions, etc are allowed here.
```

Only when we statically mention the cluster name in the `moved` block we get 0 destroy.
Screenshot 2023-04-04 at 1 49 29 PM

cluster.tf
```diff
@@ -1,104 +1,72 @@
...
-resource "ibm_container_vpc_cluster" "cluster" {
+
+module "cluster" {
+  depends_on = [
+    module.vpc
+  ]
   for_each          = local.clusters_map
-  name              = "${var.prefix}-${each.value.name}"
-  vpc_id            = each.value.vpc_id
+  source            = "git::https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc.git?ref=v3.0.0"
```
argeiger, line 19: Doesn't look like we are passing in the ocp version

argeiger (changes requested):

cluster.tf
```diff
@@ -1,104 +1,72 @@
...
+        id         = zone.id
+        zone       = zone.zone
+        cidr_block = zone.cidr
+      }
+    ]
+  }
+  worker_pools = concat(
+    [
+      {
+        subnet_prefix    = each.value.subnet_names[0]
```
argeiger, line 39: This will fail since you are creating a map up above with the key called "vsi-zone-1".  If we are using a different subnet with a different name, it will fail.  You should make it generic and pass in the value vs utilizing the subnet name from the configuration to index vpc_subnets.

Example: Using vsi-zone-2 
```│ Error: Insufficient zones blocks
│ 
│   on .terraform/modules/landing_zone.cluster/main.tf line 51, in resource "ibm_container_vpc_cluster" "cluster":
│   51: resource "ibm_container_vpc_cluster" "cluster" {
│ 
│ At least 1 "zones" blocks are required.
╵
╷
│ Error: Invalid index
│ 
│   on .terraform/modules/landing_zone.cluster/main.tf line 77, in resource "ibm_container_vpc_cluster" "cluster":
│   77:     for_each = var.vpc_subnets[local.default_pool.subnet_prefix]
│     ├────────────────
│     │ local.default_pool.subnet_prefix is "vsi-zone-2"
│     │ var.vpc_subnets is map of list of object with 1 element
│ 
│ The given key does not identify an element in this collection value.
```

argeiger (changes requested):

cluster.tf
```diff
@@ -1,104 +1,72 @@
...
+        machine_type     = each.value.machine_type
+        workers_per_zone = each.value.workers_per_subnet
+        boot_volume_encryption_kms_config = {
+          crk             = module.key_management.key_map[each.value.kms_config.crk_name].key_id
+          kms_instance_id = module.key_management.key_management_guid
+        }
+      }
+    ],
+    [
+      for pool in each.value.worker_pools :
```
argeiger, line 50: Since "worker_pools" is optional, can you do a lookup on the value and set to [] if it is null?

argeiger (changes requested):

argeiger:
I deployed it and doesn't look like we have access to the master nodes - 

image

ACLs
image

@vburckhardt

vburckhardt:
There is a related issues in this area https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/393

transit_gateway.tf
```diff
@@ -1,3 +1,11 @@
+data "ibm_is_vpc" "example" {
```
vburckhardt, line 1: Are the changes to transit_gateway.tf attempting to fix another unrelated issue? Is there more context?

variables.tf
```diff
@@ -755,19 +761,25 @@ variable "clusters" {
...
-      cos_name           = optional(string) # Name of COS instance Required only for OpenShift clusters
-      update_all_workers = optional(bool)   # If true force workers to update
+      name                            = string           # Name of Cluster
+      vpc_name                        = string           # Name of VPC
+      subnet_names                    = list(string)     # List of vpc subnets for cluster
+      workers_per_subnet              = number           # Worker nodes per subnet.
+      machine_type                    = string           # Worker node flavor
+      kube_type                       = string           # iks or openshift
+      kube_version                    = optional(string) # Can be a version from `ibmcloud ks versions` or `latest`
+      logdna_plan                     = optional(string) # Logging plan to provision
```
vburckhardt, line 771: How are the logdna_plan and sysdig_plan used here given that the ocp module does not deploy logdna/sysdig?

variables.tf
```diff
@@ -755,19 +761,25 @@ variable "clusters" {
...
+      sysdig_plan                     = optional(string) # Monitoring plan to provision
+      disable_public_endpoint         = optional(bool)   # Flag indicating that the public endpoint should be disabled
+      verify_worker_network_readiness = optional(bool)   # Flag to run a script will run kubectl commands to verify that all worker nodes can communicate successfully with the master. If the runtime does not have access to the kube cluster to run kubectl commands, this should be set to false.
+      enable_platform_logs            = optional(bool)   # Receive platform logs in the provisioned IBM Cloud Logging instance.
+      enable_platform_metrics         = optional(bool)   # Receive platform metrics in the provisioned IBM Cloud Monitoring instance.
+      entitlement                     = optional(string) # entitlement option for openshift
```
vburckhardt, line 777: Suggest to keep the same variables as in the original here to simplify.

patterns/roks/variables.tf
```diff
@@ -163,6 +163,42 @@ variable "entitlement" {
...
+  default     = true
+}
+
+variable "verify_worker_network_readiness" {
+  type        = bool
+  description = "By setting this to true, a script will run kubectl commands to verify that all worker nodes can communicate successfully with the master. If the runtime does not have access to the kube cluster to run kubectl commands, this should be set to false."
+  default     = false
+}
+
+variable "logdna_plan" {
```
vburckhardt, line 178: I don't think the code is setting up anything related with logdna / sysdig here.

.docs/pattern-defaults.md
```diff
@@ -202,7 +202,7 @@ Clusters deployed use the most recent default cluster version.
 
 ### Workers per zone
 
-The number of workers in each zone of the cluster can be changed by using the `workers_per_subnet` variable. At least two workers must be available for clusters to successfully provision.
+The number of workers in each zone of the cluster can be changed by using the `workers_per_zone` variable. At least two workers must be available for clusters to successfully provision.
```
vburckhardt, line 205: I think we should be able to stick with the original `workers_per_subnet` variable

cluster.tf
```diff
@@ -1,104 +1,78 @@
-##############################################################################
-# Find valid IKS/ROKS Cluster versions for region
-##############################################################################
-
-data "ibm_container_cluster_versions" "cluster_versions" {}
-
-##############################################################################
-
-
 ##############################################################################
```
vburckhardt, line 1: The logic related to creating iks cluster should remain here. The module should be used for ocp clusters only

cluster.tf
```diff
@@ -1,104 +1,78 @@
...
-  disable_public_service_endpoint = true
 
-  timeouts {
-    create = "3h"
-    delete = "2h"
-    update = "3h"
-  }
-
-}
+# locals {
```
vburckhardt, line 14: Remove comment

cra-tf-validate-ignore-goals.json
```diff
@@ -36,6 +36,18 @@
             "description": "Check whether Flow Logs for VPC are enabled",
             "ignore_reason": "Flow logs are enabled, so seems to be CRA bug. Tracking in https://github.ibm.com/oneibmcloud/CD-CRA/issues/1779",
             "is_valid": false
+        },
+        {
```
vburckhardt, line 40: Are those new warnings when moving to the module? There is no difference in this area between what the module does and the original version, so surprised.

vburckhardt (changes requested):
Hi @Aashiq-J - couple of additional comments

vburckhardt:
On the backward compatibility aspect - it is possible to move forward to this version without destroying the cluster by doing as follow. This would be easily scriptable:

```
1. terraform state list | grep "ibm_container_vpc_cluster.cluster"
module.landing_zone.ibm_container_vpc_cluster.cluster["debug-ocp-management-cluster"] 
module.landing_zone.ibm_container_vpc_cluster.cluster["debug-ocp-workload-cluster"] 
2. Take the cluster name from the list - eg:
debug-ocp-management-cluster
debug-ocp-workload-cluster
3. For each element in the list:
terraform state mv module.landing_zone.ibm_container_vpc_cluster.cluster["<cluster name>"] module.landing_zone.module.cluster["<cluster name"].ibm_container_vpc_cluster.cluster[0]
```

I think we should discuss / brainstorm about the best way to package this logic for the consumer. Technical consumer can run the script directly during upgrade - but this won't work for users using the logic from IBM Cloud Schematics. 
Ideally this would come as a pre-apply run (terragrunt has got that concept). Perhaps using a data external block to run the script but this may have deep side effects to evaluate further as changing the state during plan evaluation.

transit_gateway.tf
Aashiq-J: This logic was added as part of deploying management and workload vpc separately to create a fully private cluster during with public endpoint disable, but with the new base-ocp code which disables `confirm_network_healthy` when public endpoint is disabled, this change is deprecated. Will remove it.

variables.tf
Aashiq-J: It was added as part of the earlier change when we deployed logdna & sysdig, but the plans were later drops. Kept the variable thinking it will be introduced in later phases. Hence kept them optional.

ocofaigh:
FYI, IBM schematics has a terraform state mv command too -> https://cloud.ibm.com/docs/schematics?topic=schematics-schematics-cli-reference#schematics-wks_statemv
`ibmcloud schematics workspace state mv --id WORKSPACE_ID --source SOURCE  --destination DESTINATION`

vburckhardt:
@ocofaigh - Right, most terraform cli commands are exposed through ibmcloud cli schematics. This issue here is not the commands themselves, but how do we get non-tech savvy people to run those automatically - eg: there is no equivalent of terragrunt hooks in schematics afaik.

cra-tf-validate-ignore-goals.json
vburckhardt: Do we know more on this?

variables.tf
vburckhardt: pod_subnet and service_subnet are still defined here - but they are not used in the code from a quick glance.

cluster.tf
```diff
@@ -79,11 +80,14 @@ resource "ibm_container_vpc_cluster" "cluster" {
 
 
 ##############################################################################
-# Create Worker Pools
+# Create IKS Worker Pools
```
vburckhardt, line 83: I'd suggest to add at least 1 test that exercise this logic.

vburckhardt (commented):
Hi @Aashiq-J - some more comments. Thanks for pushing on this, I think it's progressing towards the right direction.

cra-tf-validate-ignore-goals.json
ocofaigh: Looks like `3000408` is no longer failing using latest version of CRA so can be removed from this

ocofaigh: Looks like `3000452`, `3000805` and `3000804` are also not failing either and should be removed

ocofaigh: PS: Soon we will move to running CRA with --sccv2 profile, so I imagine the contents of this JSON will change

Aashiq-J:
Hi @vburckhardt,
the pod_subnet and service_subnets are optional inputs for the iks cluster creation.
Screenshot 2023-04-25 at 7 12 32 PM

Aashiq-J:
@SirSpidey, @vburckhardt 
Please review this documentation change for preventing re-creation of cluster during the major version change.
https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/pull/253/commits/26a41db0bd7bcd06c9b5347e78c4d33f5ccf6335

README.md
```diff
@@ -67,6 +67,48 @@ Complete the following steps before you deploy the Secure Landing Zone module.
...
+1. terraform state list | grep "ibm_container_vpc_worker_pool.pool"
+module.landing_zone.ibm_container_vpc_worker_pool.pool["debug-ocp-workload-cluster-logging-worker-pool"]
+
+2. Seperate the cluster name and the worker pool name. For example:
+cluster_name : debug-ocp-workload-cluster
+worker_pool_name : logging-worker-pool
+
+3. For each element in the list, Run the following:
+terraform state mv 'module.landing_zone.ibm_container_vpc_worker_pool.pool["<cluster_name>-<worker_pool_name>"]' 'module.landing_zone.module.cluster["<cluster_name>"].ibm_container_vpc_worker_pool.pool["<worker_pool_name>"]'
+```
```
ocofaigh, line 110: These steps are only going to be applicable for users upgrading the roks pattern from version A -> B, so putting them in the main readme may cause confusion. Perhaps they need to go into their own .md file (maybe in the .docs folder) and then a link to them included in the release notes for the new version.

Also we need to mention something about schematics users.

SirSpidey: I agree with Conall about putting this in its own file and linking.

README.md
```diff
@@ -67,6 +67,48 @@ Complete the following steps before you deploy the Secure Landing Zone module.
 
     Select **All Identity and Access enabled services** when you assign access to the group.
 
+### Steps to prevent Destroy and recreation of cluster.
```
SirSpidey, line 70: If in it's own file, the title would probably be something like
`Updating to another version of the Red Hat OpenShift pattern`

README.md
```diff
@@ -67,6 +67,48 @@ Complete the following steps before you deploy the Secure Landing Zone module.
...
+1. Prior to `terraform plan` and `terraform apply`. Edit `moved_config.tf`
```
SirSpidey, line 71: Before jumping into the steps, users will need some brief explanation of what these steps accomplish for them. Something like,

`You can update to a new version of the Red Hat OpenShift pattern without destroying your cluster.`

README.md
```diff
@@ -67,6 +67,48 @@ Complete the following steps before you deploy the Secure Landing Zone module.
...
+2. Add the following lines to the terraform file.
```
SirSpidey, line 72: which file, `/patterns/roks/main.tf`?

README.md
```diff
@@ -67,6 +67,48 @@ Complete the following steps before you deploy the Secure Landing Zone module.
...
+  moved {
+    from = ibm_container_vpc_cluster.cluster["<workload_cluster_name>"]
+    to   = module.cluster["<workload_cluster_name>"].ibm_container_vpc_cluster.cluster[0]
+  }
+  moved {
+    from = ibm_container_vpc_cluster.cluster["<management_cluster_name>"]
+    to   = module.cluster["<management_cluster_name>"].ibm_container_vpc_cluster.cluster[0]
+  }
+  ```
+3. Make sure to replace the `<workload_cluster_name>` with the actual workload cluster name. Similarly `<management_cluster_name>` with the management cluster name.
```
SirSpidey, line 83:
```suggestion
    Where `<workload_cluster_name>` and `<management_cluster_name>` are your workload cluster and management cluster names.
```

README.md
```diff
@@ -67,6 +67,48 @@ Complete the following steps before you deploy the Secure Landing Zone module.
...
+4. Go ahead with `terraform plan` and `terraform apply`.
+
+#### Alternative approach using terraform cli:
```
SirSpidey, line 86: Because there are alternatives, I suggest you list them after the opening description. Something like, 

Choose the method that matches how you work with the module: 
- Modify the Terraform code
- Use the Terraform CLI
- Use Schematics

Modify this so that it makes sense though.

README.md
SirSpidey, line 110: If it's a set of commands to run, introduce each (or the set), but don't put numbers with the commands. Like, 

1. Get the list of Terraform state files by running this command: 
    ```tf
    terraform state list | grep "ibm_container_vpc_worker_pool.pool" module.landing_zone.ibm_container_vpc_worker_pool.pool["debug-ocp-workload-cluster-logging-worker-pool"]
    ```

1.  Locate the cluster and worker pool names. (not sure why you're doing this. can you provide more context?)
1.  Run the following command for each element in the list

README.md
```diff
@@ -679,6 +721,8 @@ list(
 
 You can create as many `iks` or `openshift` clusters and worker pools on VPC. For `ROKS` clusters, make sure to enable public gateways to allow your cluster to correctly provision ingress application load balancers.
 
+NOTE: The cluster created will have a single worker pool with worker nodes distributed across 3 different subnet each in a different availability zone.
```
SirSpidey, line 724: - Will everyone understand why this is important? 
- Does this belong with HPCS?

Also, I'm not sure that it needs a `Note`. I think it could be just a paragraph.

SirSpidey (changes requested):
some questions and suggestions

README.md
Aashiq-J: It could depend, it can be the /patterns/roks/main.tf or maybe the /examples/one-vsi-one-roks/main.tf

README.md
Aashiq-J: This is no longer the case. This statement can be removed.

SirSpidey:
@Aashiq-J I created a PR #487 for the new readme that merges into this pr. If the content makes sense, merge it and then we can continue with this PR.

ocofaigh:
If we get this working end to end, it would be a great thing to show on next weeks playback or derp dive

PR #284, opened 2023-03-01
docs: use templating to auto set version, etc. within usage

kbiegert:
### Description

Use more template variables for the `usage`.  Remove catalogId spec from usage because with it specified the url will require access to the private catalog instead of the ibm public catalog which this offering is published to for a set of allowlisted accounts.

ROKS is not yet published that I am aware of so it remains with a catalogId.


kbiegert:
When you get a chance for a review @ocofaigh

ocofaigh (approved):

Issue #355, opened 2023-04-04
Papercut: Lookup for ssh_public_key if the one passed as input already exists in account / region


vburckhardt:
### Description

This is a recurring issues with many consumers where the ssh_public_key that is passed as input already exists in the account. Backend returns with an error stating that key already exists. 

Investigate whether we could do a better job at the terraform layer by looking up the existing keys in account and using https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/data-sources/is_ssh_keys and comparing with the key passed as input.

imprateeksh:
Closing the issue as PR is merged.

PR #434, opened 2023-05-08
fix: failure to delete newly created ssh keys 

imprateeksh:
### Description
This change is to address the regression occurred as part of [papercut issue - 355](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/355)

The regression was causing the newly created SSH key not to be deleted but with the changes added this issue is addressed.

#### No release required

- [ ] Documentation update
- [ ] Other changes that don't affect Terraform code

#### Release required

- [x] Bug fix (patch release (`x.x.X`): Change that fixes an issue and is compatible with earlier versions)
- [ ] New feature (minor release (`x.X.x`): Change that adds functionality and is compatible with earlier versions)
- [ ] Breaking change (major release (`X.x.x`): Change that is likely incompatible with previous versions)

##### Release notes content

In version v3.5.0, a change was made to reuse an existing SSH key if the key passed as input already existed. However, this change caused a regression that was discovered later. Although the SSH key resources were successfully provisioned, it was observed that the SSH key resource was not being deleted as expected.

**What caused the failure in deleting the SSH key?**

 - The failure in deleting the SSH key occurred due to a change in the newly introduced flag value, which was used to determine whether to create an SSH key resource or not.
 - Once the SSH key resource was created, this flag was modified to false so that the SSH key would not be recreated (since it already existed in the VPC).
 - During the next `terraform plan` after running `terraform apply`, Terraform discovered that the resource needed to be deleted because the flag value prevented it from creating a new resource or entering the resource block.
 - However, the state file still contained the SSH key resource information that was created during the first execution of `terraform apply`.
 - Therefore, the change was made to address the above situation.
---

### Checklist for reviewers

### Merge actions for mergers

- Use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents and any release notes provided by the PR author.

    The commit message determines whether a new version of the module is needed, and if so, which semver increment to use (major, minor, or patch).


imprateeksh:
### **(1) Passing ssh key that is not present by Variable (for all three patterns)**  

**(a) Mix Pattern: (Using 4096 bits key)**  
Apply successful, Plan after Apply also successful, Destroy successful
image

**(b) VSI Pattern: (Using 2048 bits key - simply chosen to see the results for 2048 bit SSH key)**  
Apply successful, Plan after Apply also successful, Destroy successful

**(c) ROKS Pattern: (No key was provided ; this is already present in ROKS)**  
Apply successful, Plan after Apply also successful, Destroy successful

### **(2) Passing ssh key that already exists by Variable (for two patterns)**
**(a)	Mix Pattern:** 
Apply – Successful, Plan after Apply - Successful, Destroy - Successful

**(b)	VSI Pattern:**
Apply - Successful, Plan after Apply - Successful, Destroy - Successful

(c) ROKS Pattern : SSH Key is not used even though provided as no vsi is created here, hence skipped this test

### **(3) Passing Two ssh keys (1 new and 1 already existing) using `override.json` file  (for two patterns)**
**(a) Mix Pattern:**  (Currently In progress)
Apply –  , Plan after Apply -, Destroy – 

**(b)  VSI Pattern:**
Apply –  Successful, Plan after Apply - Successful, Destroy – Successful
 
image

**(c)  ROKS Pattern:**
Skipped ROKS, unrelated to SSH key currently even if it is provided.

### **(4) Passing two ssh key (1 new and 1 already existing) using `override` string (for one pattern)**

**(a) MIX Pattern** - Skipped this (time taking)

**(b) VSI**
Apply – successful ; Plan after apply – successful ; Destroy - Successful
image

**(c) ROKS** - Skipped this test

### **(5)**	Error is thrown when expected key is not found in the selected region during the plan phase (Expected)
image

**Some logs :**
[vsi_apply_using_override_json.txt](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/files/11424019/apply_using_override.txt)
[vsi-override-json-resources-list.txt](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/files/11424134/vsi-override-json-resources-list.txt)
[vsi-using-override-string.txt](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/files/11424135/vsi-using-override-string.txt)

imprateeksh:
Getting the below error in multiple runs for mix pattern when run using override.json file
One instance of error itself takes > 2hrs. I tried 3-4 times. Also tried changing DNS entries to see if the below error gets resolved

image

patterns/vsi/config.tf
```diff
@@ -46,6 +46,19 @@ locals {
   }
   override_type = var.override_json_string == "" ? "override" : "override_json_string"
 
+
+  ##############################################################################
+  # Default SSH key
+  ##############################################################################
+
+  ssh_keys = [
```
toddgiguere, line 54: Was there a reason why you moved this block out of "local.config" block? For consistency I think it still belongs under the "local.config" as that is where all the other defaults are configured for the pattern.

toddgiguere (changes requested):

toddgiguere:
Overall I think this looks good, although haven't tested it yet. One other thing to add would be a validation that the ssh_key variable and existing variable cannot both be "null" for VSI or mixed pattern, and probably also if they are both supplied? Terraform doesn't let you do a regular input validation based on another input variable, so we would need to do that with the special "regex" workaround, example here: https://github.com/terraform-ibm-modules/terraform-ibm-base-ocp-vpc/blob/main/main.tf#L21

patterns/vsi/config.tf
imprateeksh: Yes Todd, If this was not moved, I was getting below error during plan phase itself: 
`Error: Local value local.config cannot use its own result as part of its expression.`

imprateeksh:
I have made the change using `tobool()` for **validation 1** above . This worked fine when checked with assigning values using variables. 
Now I am validating using override.json

imprateeksh:
The example I tried using override.json worked successfully.

ocofaigh:
Great news, @toddgiguere Can you give this a final review please?

PR #435, opened 2023-05-08
feat: remove rules from default VPC security group and ACL

toddgiguere:
### Description

Allows the use of new feature in the landing-zone-vpc-module to remove (clean) all rules from the default security groups and ACLs that get created with a VPC.

Currently, IBM Cloud does not allow removal of the default security group or ACL that are automatically created with a VPC, and the default rules are very permissive. Since secure landing zone creates its own groups, the defaults are not needed. They cannot be removed in whole, but the rules can be completely removed from both defaults to increase security posture.

#### No release required

- [ ] Documentation update
- [ ] Other changes that don't affect Terraform code

#### Release required

- [ ] Bug fix (patch release (`x.x.X`): Change that fixes an issue and is compatible with earlier versions)
- [ ] New feature (minor release (`x.X.x`): Change that adds functionality and is compatible with earlier versions)
- [x] Breaking change (major release (`X.x.x`): Change that is likely incompatible with previous versions)

##### Release notes content

All patterns will now utilize the `clean_default_security_group` and `clean_default_acl` variables added to `terraform-ibm-landing-zone-vpc` in [v7.1.0](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone-vpc/releases/tag/v7.1.0), which if set to "true" will remove all rules from the default security groups and ACLs created automatically when new VPCs are deployed.

All patterns will now set both variables to "true" by default for all VPCs, and you can override these values for each VPC when using the "override.json" file.

Since the default security group and ACL for a VPC are created with very permissive rules, setting these values to "true" to remove those rules will increase the overall security posture of the deployment.

---

### Checklist for reviewers

### Merge actions for mergers

- Use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents and any release notes provided by the PR author.

    The commit message determines whether a new version of the module is needed, and if so, which semver increment to use (major, minor, or patch).


ocofaigh (dismissed):
LGTM

ocofaigh (dismissed):

vburckhardt:
I think the release note should includes the following details for consumers:
1. How to migrate an existing override.json to be compatible with new version (eg: adding clean_default_security_group and clean_default_acl )
2. How to disable the new behavior that deletes network acl rules and sg rules.

ocofaigh:
@toddgiguere can you help get release notes updated before we merge? Thanks

toddgiguere:
>     2. How to disable the new behavior that deletes network acl rules and sg rules.

The only way to disable the behavior in the patterns is to use the override.json. I had figured that since the patterns do not attach default ACL and do not include default security group rules, the patterns will set these to "true" always (opinionated).

ocofaigh (approved):

PR #455, opened 2023-06-06
feat: Expose HPCS support in IBM Cloud Catalog releases

daniel-butler-irl:
### Description

Expose HPCS support and add tests

#### Changes that affect the core Terraform module or submodules
- [ ] Bug fix
- [x] New feature
- [ ] Dependency update

#### Changes that don't affect the core Terraform module or submodules
- [ ] Documentation update
- [ ] Other

#### Release required?
Identify the type of release. For information about the changes in a semantic versioning release, see [Release versioning](https://terraform-ibm-modules.github.io/documentation/#/versioning).

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

##### Release notes content

If a release is required, replace this text with information that users need to know about the release. Write the release notes to help users understand the changes, and include information about how to update from the previous version.

Your notes help the merger write the commit message for the PR that is published in the release notes for the module.

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers

- Use a relevant [conventional commit](https://www.conventionalcommits.org/) message that is based on the PR contents and any release notes provided by the PR author. The commit message determines whether a new version of the module is needed, and if so, which semver increment to use (major, minor, or patch).


ocofaigh:
Some comments..

Can we just update the `TestRunRoksPattern` and `TestRunVsiPattern` tests to use HPCS and then leave the `TestRunVpcPattern` to use Key Protect? 

Can you please update the `catalogValidationValues.json.template` file in all 3 patterns to now pass the HPCS variables?

We need to do an SCC scan using the new HPCS variables to see if the rule `Check whether Cloud Object Storage is enabled with customer-managed encryption and Keep Your Own Key (KYOK)` is now passing. If it does, we may be able to claim more controls in ibm_catalog.json. Affected controls are `SC-13(0)`, `SC-28(0)` and `SC-28(1)(0)`

daniel-butler-irl:
confirmed `Check whether Cloud Object Storage is enabled with customer-managed encryption and Keep Your Own Key (KYOK)` passes SCC scan now

ibm_catalog.json
```diff
@@ -668,6 +666,14 @@
                             "key": "region",
                             "required": true
                         },
+                        {
```
daniel-butler-irl, line 669: @ocofaigh this is supposed to be removed right? I think it got back in via a merge resolve

ocofaigh: yep, we no longer need to explicitly specify required variables - catalog now marks them as required if they dont have a default value in terraform

ocofaigh (dismissed):

ocofaigh (dismissed):
LGTM

ocofaigh:
Oh @daniel-butler-irl We have a problem now. The test is failing with:
```
Error: [ERROR] Error creating authorization policy: The policy wasn't created because an access policy with identical attributes already exists. Please update the roles in the existing policy (50a70904-62fe-405c-b555-485e32faf016), or update the one you're trying to assign to include a different attribute assignment.
```
This is because we are using our permanent HPCS instance, so there may already be a policy in our account. By any chance is our code actually trying to create duplicate auth policies? Or do we actually have something lingering in our account?

daniel-butler-irl:
I'll need to dig into this on Monday

daniel-butler-irl:
This prevents us from running hpcs tests in parallel. https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/main/dynamic_values/config_modules/service_authorizations/service_authorizations.tf#L43
I removed the parallel setting in and will create an internal issue to track a fix.

vburckhardt:
I think this needs to be coordinated with @SirSpidey in term of documentation - eg: at minimum update the section https://cloud.ibm.com/docs/secure-infrastructure-vpc?topic=secure-infrastructure-vpc-plan#vpc-crypto-prereqs to mention the fields.

(one issue I just noticed - we already have the hpcs section documented as pre-req, but no way to specify the hpcs instance yet until this is released )

vburckhardt (approved):

Issue #510, opened 2023-07-27
Support for existing encryption keys referenced by CRN


in-1911:
### Description

Enable support for using an existing key referenced by CRN in the override.json (`key_management.keys` element).
This will allow to use a cross-account KeyProtect / HPCS Instances where there may be no full visibility to the KMS instance, but the keys are available by CRN.

- If a CRN of a key is provided, no KMS instance should be required to look up the key, the key CRN is just included in the key map to be used in other modules referencing it by name
- The landing zone should support a scenario where no KMS instance is supplied but a set of existing keys with CRNs is provided in override.json (`key_management.keys` element)
- The landing zone should support a mixed scenario where some keys are supposed to be created while some are provided with CRNs.

### New or affected modules

- kms/variables.tf - add crn attributes as optional(string) to the variable schema
- kms/outputs.tf - key_map should be a union of keys created with references of existing keys that CRNs were provided for
- kms/dynamic_values/main.tf - key_management_keys, key_rings, and key_management_key_policies should only include items for which keys would be actually created. Key items with existing CRNs are omitted from those lists.

---

By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/terraform-ibm-modules/documentation/blob/main/CODE_OF_CONDUCT.md)

Aashiq-J:
@in-1911 ,
When a kms instance is not provided slz goes ahead and creates the kms instance as it is required by the cluster as we need to attach a Key Protect instance to the cluster.
https://registry.terraform.io/providers/JonWoodlief/ibm/latest/docs/resources/container_vpc_cluster#create-a-kms-enabled-kubernetes-cluster

So, I think its not possible to not pass a kms instance, but we can use existing kms instance or create a new one.
I can go ahead and implement the feature of using existing key crn.

ocofaigh:
@in-1911 Looking at the api (https://cloud.ibm.com/apidocs/kubernetes/containers-v1-v2#createkmsconfig) it shows required for `instanceID` and `crkID`, however it does support passing `accountID` which is supposed to be used if the KMS instance is in another account. We have created an enhancement request with the IBM terraform provider to see if they will add support to pass an account ID for this use case -> https://github.com/IBM-Cloud/terraform-provider-ibm/issues/4745

ocofaigh:
[...]

PR #542, opened 2023-08-21
feat: ability to pass external key crn for vsi, cos and secret manager

Aashiq-J:
### Description

https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/510

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers


in-1911:
The implementation seems to be somewhat complicated (now the key CRNs will have to be sprinkled all over the config using different names in each element) and is a bit different from the request in #510 
The idea was to allow to pass the key CRN in the key_management config so that other modules can still refer to that key by "internal alias" but the kms module will skip creating those keys since the CRNs are already provided and the output of the kms module would be a union of pre-existing keys and the keys created (if any).

PR #546, opened 2023-08-23
fix: override json

jor2:
### Description

Value was being overwritten.
https://github.ibm.com/goldeneye/issues/issues/5669

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers

### DIFF
--- a/dynamic_values/config_modules/vsi/vsi.tf
+++ b/dynamic_values/config_modules/vsi/vsi.tf
@@ -63,7 +63,7 @@ module "vsi_image_map" {
 ##############################################################################
 
 output "vsi_map" {
-  description = "Map of VSI deplioyments"
+  description = "Map of VSI deployments"
   value       = module.composed_vsi_map.value
 }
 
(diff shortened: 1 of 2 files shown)


jor2:
A lot of inputs are rendered useless by being overwritten by `override_json_string`. Should we change some of the functionality surrounding this?

ocofaigh:
@toddgiguere Any idea what happened here?

toddgiguere:
That key array in the VSI section tells the VSI which key to use for authentication. The "ssh-keys" string inside the array tells the VSI instance to use the ssh-key with the same identifier that is in another section of the override, that's how it matches the two things up.

By setting the VSI array to empty in the override, will it use the correct key for VSI?

Have we done some testing both ways? (new key vs existing key) Verify that the VSI instances are using the correct key? There is a way in the GUI to verify which auth key is associated with a VSI instance.

PR #553, opened 2023-08-29
feat: use existing KMS keys referenced by CRNs

in-1911:
### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [X] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

##### Release notes content

NOTE: 
This is my own implementation of  #510 based on some work by @Aashiq-J.
I was originally looking at #542 by @Aashiq-J and missed the later alternative draft in #547, so this PR is much more in line with #547 and provides a simpler way of managing the keys in the same place as before. I also avoided changing some other modules that use key mappings, so it's more transparent. Also the case where a KMS instance is not used at all (all keys are centrally managed) is supported a bit better.

#### Feature to use existing keys referenced by CRN possibly from a cross-account KMS instance

The keys are still configured entirely in the `key_management` element of override.json:

```json
{
  "key_management": {
    "access_tags": [],
    "keys": [
      {
        "key_ring": "slz-slz-ring",
        "name": "slz-vsi-volume-key",
        "root_key": true
      },
      {
        "name": "sample",
        "crn": "xxxx-xxxx-xxxx-xxxx"
      }
    ],
    "name": "slz-kms",
    "resource_group": "slz-service-rg",
    "use_hs_crypto": false
  }
}
```
- If a key has a `crn` attribute, other attributes are ignored and the CRN is used in any place where the key is referenced by name (using the same module parameters as before)
- If a key does not have a `crn` attribute, the key is provisioned using the KMS instance defined by attributes like `key_management.name`, `key_management.use_hs_crypto`, etc
- If `key_management.name` is not supplied, all key entries are expected to have a CRN. In that case:
  - No KMS instance is created
  - No S2S authorization policy targetting `kms` or `hs-cryptio` service is created
- A mixed scenario is supported as depicted in the above example - some keys may be supplied with CRNs while some will be provisioned in the designated KMS instance
- KMS module unit tests are updated to reflect the changes in logic
- KMS module outputs are not changed
- No other modules (COS / VSI / etc) require changes

#### Prerequisites for cross-account KMS
- Existing KMS instance (KeyProtect or HPCS) in the other account
- Keys are created and CRNs are made available
- S2S authorizations for cross account access are provisioned in the other account

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

- [X] If relevant, a test for the change is included or updated with this PR.
- [X] If relevant, documentation for the change is included or updated with this PR.

### Merge actions for mergers


in-1911:
@Aashiq-J - could you please run pipeline on this one?

in-1911:
Recommended changes have been implemented in #547 , closing this one

PR #563, opened 2023-09-04
test: Add override test

akocbek:
### Description

Get config output value from first apply and use the value as input variable (override json) on second apply. After check consistency

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers


tests/pr_test.go
```diff
@@ -216,3 +219,60 @@ func TestRunUpgradeVpcPattern(t *testing.T) {
...
+	options.SkipTestTearDown = true
+	output, err := options.RunTestConsistency()
+
+	if assert.Nil(t, err, "This should not have errored") &&
+		assert.NotNil(t, output, "Expected some output") &&
+		assert.NotNil(t, options.LastTestTerraformOutputs, "Expected some Terraform outputs") {
+		// set override json string with previous value of config output
+		options.TerraformOptions.Vars["override_json_string"] = options.LastTestTerraformOutputs["config"]
+
+		options.SkipTestSetup = true
```
toddgiguere, line 236: Now that you are not using wrapper for 2nd plan, you don't need to set SkipTestSetup.

tests/pr_test.go
```diff
@@ -216,3 +219,60 @@ func TestRunUpgradeVpcPattern(t *testing.T) {
...
+		// TERRATEST uses its own internal logger.
+		// The "show" command will produce a very large JSON to stdout which is printed by the logger.
+		// We are temporarily turning the terratest logger OFF (discard) while running "show" to prevent large JSON stdout.
+		options.TerraformOptions.Logger = logger.Discard
+		planStruct, err := terraform.InitAndPlanAndShowWithStructE(options.Testing, options.TerraformOptions)
```
toddgiguere, line 241: More of a golang style recommendation, I have seen issues in the past when reusing variables on the left hand side of the `:=` operator, in this case reusing the variable `err` that was used for previous function call. To avoid any runtime issues you could make sure that you use unique error variables for each call (like for instance call this one `errPlan` or similar). Sometimes golang deals with it ok, and other times it complains, I've been taking the safe route and using unique names.

tests/pr_test.go
```diff
@@ -216,3 +219,60 @@ func TestRunUpgradeVpcPattern(t *testing.T) {
...
+				}
+
+				var errorMessage string
+				errorMessage = fmt.Sprintf("Resource(s) identified to be destroyed %s", resourceDetails)
+				noChange := resource.Change.Actions.NoOp() || resource.Change.Actions.Read()
+				assert.True(options.Testing, noChange, errorMessage)
+			}
+
+			// Run plan again to output the nice human-readable plan if there are valid changes
+			if noChange {
```
toddgiguere, line 272: You probably wanted to print out human readable plan only if there are changes, so this should probably be the negative, like `!noChange`

tests/pr_test.go
```diff
@@ -216,3 +219,60 @@ func TestRunUpgradeVpcPattern(t *testing.T) {
...
+				if resource.Change.Actions.Update() {
+					resourceDetails = fmt.Sprintf("Name: %s Address: %s Actions: %s\nDIFF:\n%s\n\nChange Detail:\n%s", resource.Name, resource.Address, resource.Change.Actions, common.GetBeforeAfterDiff(changesJson), changesJson)
+				} else {
+					// Do not print changesJson because might expose secrets
+					resourceDetails = fmt.Sprintf("Name: %s Address: %s Actions: %s\n", resource.Name, resource.Address, resource.Change.Actions)
+				}
+
+				var errorMessage string
+				errorMessage = fmt.Sprintf("Resource(s) identified to be destroyed %s", resourceDetails)
+				noChange := resource.Change.Actions.NoOp() || resource.Change.Actions.Read()
```
toddgiguere, line 267: Two comments here:
1. In golang, the `:=` operator means to assign the value to a new variable in memory, and `=` means to assign value to an existing variable. Since you have a `noChange := true` above, and now using `noChange := ` inside this for loop, this might not be doing what you intended.
2. The logic of `noChange` value might be off here. The `assert` line that follows will not quit the test, only mark it as failed, and then move on with the rest of the `for` loop. So if you want `noChange` to keep track of overall it might not work, since the value will be whatever the last resource was in the loop, and not overall. Probably need one variable to keep track of overall change, and another for the current resource change, and manage so that the `noChange` variable outside the loop indicates that every resource had no change, or "at least one" did have a change.

toddgiguere (changes requested):

tests/pr_test.go
```diff
@@ -216,3 +219,68 @@ func TestRunUpgradeVpcPattern(t *testing.T) {
...
+					// Do not print changesJson because might expose secrets
+					resourceDetails = fmt.Sprintf("Name: %s Address: %s Actions: %s\n", resource.Name, resource.Address, resource.Change.Actions)
+				}
+
+				// build error message
+				var errorMessage string
+				errorMessage = fmt.Sprintf("Resource(s) identified to be destroyed %s", resourceDetails)
+
+				// check if current resource is changed
+				currentResourceChanged := resource.Change.Actions.NoOp() || resource.Change.Actions.Read()
```
toddgiguere, line 270: If NoOp and Read are true, that means there was NOT a resource change. This line and the assert are correct, the variable name is wrong.

However, I think the if condition on line 274 is actually wrong, probably because of the way the variable reads. If "currentResourceChanged = true" that actually means not changed, which means the condition to change "resourcesChanged" is incorrect?

toddgiguere (changes requested):

toddgiguere (approved):

ocofaigh (approved):

PR #595, opened 2023-10-04
feat: add pattern for vsi creation on existing vpc

Aashiq-J:
### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### Merge actions for mergers


ocofaigh:
I'm not sure if this is exactly as we discussed in the last deep dive. But perhaps we jump on a call with @toddgiguere to discuss?

patterns/dynamic_values/config_modules/security_groups/security_groups.tf
```diff
@@ -151,11 +151,10 @@ output "f5_bastion" {
...
   description = "Security Group for Bastion VSI"
   value = {
-    name           = "bastion-vsi-sg"
-    vpc_name       = var.bastion_vpc_name
-    resource_group = "${var.prefix}-${var.bastion_vpc_name}-rg"
-    rules          = var.bastion_vsi_rules
-    access_tags    = []
+    name        = "bastion-vsi-sg"
+    vpc_name    = var.bastion_vpc_name
+    rules       = var.bastion_vsi_rules
```
toddgiguere, line 156: I think it might be risky to remove resource_group here as it is used directly in the virtual_server.tf for the VSI pattern.

I think when deploying VSIs in existing VPCs, the resource group should be known (or asked for by consumer) and sent to the VSI submodule.

If you really don't care about this output, I suppose you could just pass in a known vpc_name to this input var for bastion so that it doesn't fail?

patterns/vsi-extension/config.tf
```diff
@@ -0,0 +1,90 @@
+##############################################################################
+# Create Pattern Dynamic Variables
+# > Values are created inside the `dynamic_modules/` module to allow them to
+#   be tested
+##############################################################################
+
+module "dynamic_values" {
```
toddgiguere, line 7: the `dynamic_values/config_modules/security_groups` is a module that can be called here on its own. If you can, you can call that directly here instead of the main `dynamic_values`. This will help, because many of the changes you have made to other dynamic values modules were only necessary because of the values in this special case, but if all you need is subnet config then you can call that directly and you will not have to alter all of the other modules.

patterns/vsi-extension/version.tf
```diff
@@ -0,0 +1,30 @@
+# terraform {
```
toddgiguere, line 1: probably should remove comment blocks like this.

tests/pr_test.go
```diff
@@ -395,3 +396,116 @@ func TestRunVPCPatternSchematics(t *testing.T) {
 	err := options.RunSchematicTest()
 	assert.NoError(t, err, "Schematic Test had unexpected error")
 }
+
+func TestRunSLZExample(t *testing.T) {
```
toddgiguere, line 400: This is not a good name for this test. This is a test of the vsi extension.

Also, we need to look at this test more closely in a deeper dive, as I'm worried about a few things with a test this complex, for instance that tear down of ALL resources is done properly in all scenarios, and that the test will correctly work with a PR initiated from a fork.

patterns/vsi-extension/config.tf
```diff
@@ -0,0 +1,90 @@
...
+  ##############################################################################
+
+  ##############################################################################
+  # Dynamic configuration for landing zone environment
+  ##############################################################################
+
+  config = {
+
+    ##############################################################################
+    # VSI Configuration
```
toddgiguere, line 49: we should remove big comment blocks of code like this if they are not needed

patterns/dynamic_values/variables.tf
```diff
@@ -231,11 +223,13 @@ variable "appid_resource_group" {
 variable "teleport_instance_profile" {
   description = "Machine type for Teleport VSI instances. Use the IBM Cloud CLI command `ibmcloud is instance-profiles` to see available image profiles."
   type        = string
+  default     = null
```
toddgiguere, line 226: I think you are making changes like this just so that you can use the dynamic_values module in your special pattern, but I'm not sure we should be doing that. See other comment about pointing directly to the security_group dynamic module if that is all you need, and you won't need to make these changes.

toddgiguere (changes requested):

toddgiguere (approved):

ocofaigh:
The test `TestRunVPCPatternSchematics ` failed with this: `Error: [ERROR] Error Creating network ACL rule : Provided Name (ibmflow-allow-vpc-connectivity-inbound-038c-0) is not unique`
Not related to the code in this PR, so retrying..

ocofaigh:
Going to merge this without creating a release. I'll create the release when I add this to the ibm_catalog.json

PR #626, opened 2023-11-09
feat: fetch vpc details from the workspace

Aashiq-J:
### Description

We currently have an extension DA [here](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/tree/main/patterns/vsi-extension) (not yet published to private or public catalog) which supports deploying a VSI ontop of an existing VPC.

It was brought to our attention by the catalog team that it is possible to add terraform code into the extension DA that allows it to pull values from another schematics workspace instead of the user having to enter each of the values. So when you look at our use case, we could have terraform lookup the VPC ID by doing a data lookup on the schematics workspace where the VPC was created.

This is something the PowerVS team is doing right now with their extension DA -> https://github.com/terraform-ibm-modules/terraform-ibm-powervs-infrastructure/tree/main/solutions/extension

Issue: https://github.ibm.com/GoldenEye/issues/issues/6365

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


Aashiq-J:
Screenshot 2023-11-09 at 3 39 19 PM

When user doesn't provide workspace id or vpc_id
Screenshot 2023-11-09 at 3 43 05 PM

When user doesn't provide the vpc_name when workspace id is set.
Screenshot 2023-11-09 at 3 44 21 PM

patterns/vsi-extension/version.tf
```diff
@@ -6,9 +6,10 @@ terraform {
   required_version = ">= 1.3, < 1.6"
   # Pin to the lowest provider version of the range defined in the main module's version.tf to ensure lowest version still works
   required_providers {
+    # Pin to 1.58.1 due to https://github.com/IBM-Cloud/terraform-provider-ibm/issues/4907
     ibm = {
       source  = "IBM-Cloud/ibm"
-      version = "1.59.0"
+      version = "1.58.1"
```
ocofaigh, line 12: FYI, renovate is going to update this back to the latest due to the rules at https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/workspace/renovate.json

Are we sure that this only happens in 1.59.0 and not in 1.58.1 ?

ocofaigh (commented):
- We need to add some details to the `patterns/vsi-extension/README.md` which explains that the consumer can use the `prerequisite_workspace_id` OR the `existing_vpc_name` variable, and in what cases you may wish to use one approach over the other.

- Are the `resource_group` and `region` variables still required if using the `prerequisite_workspace_id` approach? We should review all variables now, and ensure we have an accurate and clear variable descprtion as to what is required when using either approach. 

- I notice the PowerVS guys have anoutout like this in their DAs:
```
output "schematics_workspace_id" {
  description = "ID of the IBM Cloud Schematics workspace. Returns null if not ran in Schematics."
  value       = var.IC_SCHEMATICS_WORKSPACE_ID
}
```

- I think we may need this in our VPC DA output for this to all work? Its now possible to import something to catalog using a branch, so maybe we should start testing in catalog. Ping me and we can do this together over a call.

patterns/vsi-extension/version.tf
Aashiq-J: Yes, I have created a provider issue as well for it.

Aashiq-J: https://github.com/IBM-Cloud/terraform-provider-ibm/issues/4907

ocofaigh: OK, so we need to update the renovate rules to prevent it from updating this. Its currently set up to update all of the directories in `"matchPaths": ["patterns"],` so we will need to exclude the `patterns/vsi-extension` dir

Aashiq-J:
@ocofaigh ,

- okay will update the README.

- regarding the `resource_group` variable, we can remove it and use the same resource group as the vpc to deploy the vsi. The `region` variable is required by the provider block, so its a required variable.

- the `schematics_workspace_id ` output is optional in the VPC DA, we can have it as an extra output. Our extension code will work even without this output. The user can get the workspace_id from the UI to pass it to the extension deployment.

renovate.json
```diff
@@ -21,7 +21,8 @@
       "semanticCommitType": "fix",
       "group": true,
       "groupName": "required_provider",
-      "commitMessageExtra": "to latest for the landing zone DAs"
+      "commitMessageExtra": "to latest for the landing zone DAs",
+      "ignorePaths": ["patterns/vsi-extension/"]
     }
   ]
 }
```
ocofaigh, line 28: I just noticed that the `matchPaths` array on line 7 is missing `"patterns/vsi-extension"` - can you add it please?

patterns/vsi-extension/README.md
```diff
@@ -5,3 +5,23 @@ This logic creates a VSI to an existing landing zone VPC.
...
+## Using `prerequisite_workspace_id` and `existing_vpc_name`
+
+The user can specify the workspace ID associated with the deployment of the landing zone VPC when creating a new VSI.
+
+Follow these steps to get the schematics workspace ID.
+
+1. Click the Navigation menu icon, and then click Schematics > Workspaces.
+1. Select the Workspace that is associated with landing zone VPC.
+1. Click the Settings.
+1. In the Details section, you can find the Workspace ID.
```
ocofaigh, line 24: I don't think consumer need to do this - my understanding is that when consuming from catalog, a dropdown will appear with a list of schematic workspaces, but I'm wondering if that only works if we have the `schematics_workspace_id` output. I'll start up a thread with catalog team to confirm. Either way, we will want to import this into catalog to test it. I can work with you on that

ocofaigh: Actually, sorry like you said, we already have this output in the VPC pattern, so I think what we should next is merge this code, and create an entry in our private catalog to test

ocofaigh (changes requested):
@Aashiq-J Since the pattern no longer requires `resource_group` its failing with `Error: Value for undeclared variable` - you need to update the pr test to now use `TestOptionsDefault()` instead of `TestOptionsDefaultWithVars()`

ocofaigh (approved):
Lets go with this, and create a follow up PR to add the new DA to `ibm_catalog.json` and  `.catalog-onboard-pipeline.yaml` - We can then use the branch of that PR to manually onboard it to catalog using our pipeline, where we can then do further tests before we release it. 
I'm purposely not generating a release for this PR, since we need a release to have the updated `ibm_catalog.json` in it

PR #633, opened 2023-11-20
feat: ability to fetch kms details from schematics

Aashiq-J:
### Description

Issue: https://github.ibm.com/GoldenEye/issues/issues/6661

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers

### DIFF
--- a/patterns/vsi-extension/variables.tf
+++ b/patterns/vsi-extension/variables.tf
@@ -82,8 +82,8 @@ variable "existing_kms_instance_guid" {
 
 variable "skip_iam_authorization_policy" {
   type        = bool
-  description = "Set to `true` to skip the creation of an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. If set to `false` (and creating a policy), specify the GUID of the KMS instance in the `existing_kms_instance_guid` variable."
-  default     = false
+  description = "By default (true), the Landing Zone VPC creates an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. Set to false to create the authorization policy in a different KMS instance, and specify the GUID of the KMS instance in the existing_kms_instance_guid variable."
+  default     = true
 }
 
 variable "vsi_per_subnet" {
--- a/patterns/vsi-extension/variables.tf
+++ b/patterns/vsi-extension/variables.tf
@@ -82,8 +82,8 @@ variable "existing_kms_instance_guid" {
 
 variable "skip_iam_authorization_policy" {
   type        = bool
-  description = "By default (true), the Landing Zone VPC creates an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. Set to false to create the authorization policy in a different KMS instance, and specify the GUID of the KMS instance in the existing_kms_instance_guid variable."
-  default     = true
+  description = "Set to `true` to skip the creation of an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. If set to `false` (and creating a policy), specify the GUID of the KMS instance in the `existing_kms_instance_guid` variable."
+  default     = false
 }
 
 variable "vsi_per_subnet" {
--- a/patterns/vsi-extension/variables.tf
+++ b/patterns/vsi-extension/variables.tf
@@ -84,8 +84,8 @@ variable "existing_kms_instance_guid" {
 
 variable "skip_iam_authorization_policy" {
   type        = bool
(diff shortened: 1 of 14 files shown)


patterns/vsi-extension/variables.tf
```diff
@@ -84,8 +84,8 @@ variable "existing_kms_instance_guid" {
 
 variable "skip_iam_authorization_policy" {
   type        = bool
-  description = "Set to `true` to skip the creation of an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. If set to `false` (and creating a policy), specify the GUID of the KMS instance in the `existing_kms_instance_guid` variable."
-  default     = false
+  description = "Set to `true` to skip the creation of an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. If set to `false` (and creating a policy), specify the GUID of the KMS instance in the `existing_kms_instance_guid` variable. When you are using the same KMS instance from the landing zone VPC, it may be possible to set a flag to true in order to skip creating an authorization policy since it has already been established by the VPC DA."
+  default     = true
```
ocofaigh, line 88: The wording here doesn't make sense. The description says `it may be possible to set a flag a flag to true` however the default value is true. I think the point we are trying to make is that by default the Landing Zone VPC DA will create this policy, and that is why we default it to true. Might need @SirSpidey to review how best to word this.
I'm wondering if we should be exposing it all, but I guess it is no harm to have it as an option with the default value to true so its skipped? Any thoughts on this @toddgiguere ?

I also just noticed that `existing_kms_instance_guid` has no default value, so that will make it a required variable, however it seems its only required if `skip_iam_authorization_policy` is false

SirSpidey: here's one suggestion: 

"By default (`true`), the Landing Zone VPC creates an IAM authorization policy that permits all storage blocks to read the encryption key from the KMS instance. Set to `false` to create the authorization policy in a different KMS instance, and specify the GUID of the KMS instance in the `existing_kms_instance_guid` variable."

Aashiq-J: @ocofaigh ,
even with the `skip_iam_authorization_policy ` set to false, we require the `existing_kms_instance_guid ` to encrypt the block volumes. 
Another option is to remove the default value for `skip_iam_authorization_policy ` as well, so that the user can decide whether to create the auth policy or not.

ocofaigh: @Aashiq-J OK so `existing_kms_instance_guid` is needed. Can it be looked up from schematics if user picks that approach though?

I think maybe we leave the variable in, but default it to `true` since SLZ VPC DA creates it, but lets update the description to make that clear

ocofaigh:
@Aashiq-J It looks like they fixed the provider crash you reported in https://github.com/IBM-Cloud/terraform-provider-ibm/releases/tag/v1.59.1 - do you want to try it out, and if it works we can update the DA to use it.

patterns/vsi-extension/variables.tf
Aashiq-J: @ocofaigh ,
Sorry about the confusion, `existing_kms_instance_guid ` is not required when we don't have to create an auth policy. `boot_volume_encryption_key` is the only variable required to encrypt boot volume if we skip the auth policy creation.

So, we can set null as default value for `existing_kms_instance_guid` and have a validation to check for value when `skip_iam_authorization_policy ` is set to `false`.

Aashiq-J:
@ocofaigh,
The provider bug still exists.

And I have added the ability to choose the key from the schematics.

ocofaigh (commented):
Putting my consumer hat on here, the list of variables is getting very hard to understand since we are trying to support looking up values from schematics OR passing values directly. For example, I see `boot_volume_encryption_key` and `existing_boot_volume_encryption_key_name`. I'm wondering if we need to come up with a naming convention for consumers to easily identify the variables they need to use depending on which approach they want to use. Another option could be to split this into 2 different variations. For sure we need to do better with the variable descriptions. Perhaps we set up a working session to review this next week. We could use the deep dive, or if that agenda is full, lets create an adhoc call and finally wrap this work on. Lets keep in close touch with @SirSpidey for naming on this one

ocofaigh:
Closing in favour of https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/pull/643

PR #648, opened 2023-12-04
fix: updated logic which was using deprecated lookup function

akocbek:


patterns/mixed/variables.tf
```diff
@@ -190,6 +190,7 @@ variable "entitlement" {
   default     = null
 }
 
+# tflint-ignore: terraform_unused_declarations
```
ocofaigh, line 193: why are we ignoring this? If it is not used, it should not be here, but why is it not used?

patterns/roks/module/variables.tf
```diff
@@ -136,6 +136,7 @@ variable "workers_per_zone" {
   default     = 1
 }
 
+# tflint-ignore: terraform_unused_declarations
```
ocofaigh, line 139: why are we ignoring this? If it is not used, it should not be here, but why is it not used?

variables.tf
```diff
@@ -682,6 +682,7 @@ variable "cos" {
 # Service Instance Variables
 ##############################################################################
 
+# tflint-ignore: terraform_unused_declarations
```
ocofaigh, line 685: why are we ignoring this? If it is not used, it should not be here, but why is it not used?

ocofaigh (changes requested):
@akocbek we should not be ignoring any `terraform_unused_declarations` - we need to find out why they are unused and either use them, or remove them.

patterns/mixed/variables.tf
akocbek: it is not used inside patterns, instead the value is take from override our default is set. Will remove it

variables.tf
akocbek: it is not used inside patterns, instead the value is take from override our default is set. Will remove it

patterns/mixed/variables.tf
akocbek: fixed.

patterns/roks/module/variables.tf
akocbek: fixed

Aashiq-J (approved):
Looks good to me

ocofaigh:
The `TestRunVPCPatternSchematics` test failed with:
```
2023/12/12 10:29:31 Terraform apply | Error: [ERROR] Error in getting bucket info rule: Get "https://config.cloud-object-storage.cloud.ibm.com/v1/b/ocp-sc-cmjyi2-atracker-bucket-e5pwm8i8": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```
Retrying..

PR #680, opened 2024-01-08
[DO NOT MERGE]test override

Aayush-Abhyarthi:
### Description

https://github.ibm.com/GoldenEye/issues/issues/4690

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


tests/pr_test.go
```diff
@@ -163,6 +163,57 @@ func TestRunUpgradeRoksPattern(t *testing.T) {
...
+	assert.Nil(t, err, "This should not have errored")
+	assert.NotNil(t, output, "Expected some output")
+	assert.NotNil(t, options.LastTestTerraformOutputs, "Expected some Terraform outputs")
+	assert.Nil(t, err1, "Error occured with the override json")
+	assert.Nil(t, err2, "Error occured with config json")
+	assert.Nil(t, err3, "Error occured with override_json_map")
+	assert.Nil(t, err4, "Error occured with config_map")
+	assert.True(t, jsonComp, err3)
+
+	options.TestTearDown()
```
toddgiguere, line 214: I would consider putting the TestTearDown call in a golang `defer`, right after the RunTestConsistency, so that it will be called no matter how the unit test function exits.

tests/pr_test.go
```diff
@@ -163,6 +163,57 @@ func TestRunUpgradeRoksPattern(t *testing.T) {
...
+
+	//Parse the json into a map
+	var override_json_map map[string]interface{}
+	err1 := json.Unmarshal([]byte(override_json), &override_json_map)
+	fmt.Println(err1)
+
+	config_json_byteArray, ok := config_json.([]byte)
+	if !ok {
+		fmt.Println("Error: Unable to convert config_json to []byte")
+		return
```
toddgiguere, line 185: This is a unit test, so we probably shouldn't do just a simple `return`. If you wish to fail and exit the test due to error, you can use the "github.com/stretchr/testify/require" package, which operates exactly like "assert" except if the require condition fails it will mark unit test as FAILED and also exit.

This is also another reason why you would want your TestTearDown to be registered in a golang `defer`, see below.

tests/pr_test.go
```diff
@@ -163,6 +163,57 @@ func TestRunUpgradeRoksPattern(t *testing.T) {
...
+	output, err := options.RunTestConsistency()
+
+	override_json := filepath.Join(roksPatternTerraformDir, "override.json")
+	config_json := options.LastTestTerraformOutputs["config"]
+	ignored_keys := []string{"resource_group"}
+
+	//Parse the json into a map
+	var override_json_map map[string]interface{}
+	err1 := json.Unmarshal([]byte(override_json), &override_json_map)
+	fmt.Println(err1)
```
toddgiguere, line 180: If the unmarshal has error, what should we do? is this a "FAIL and EXIT" condition? If so you can use the "require" package (see another comment below) to handle appropriately.

tests/pr_test.go
```diff
@@ -163,6 +163,57 @@ func TestRunUpgradeRoksPattern(t *testing.T) {
...
+
+	config_json_byteArray, ok := config_json.([]byte)
+	if !ok {
+		fmt.Println("Error: Unable to convert config_json to []byte")
+		return
+	}
+
+	var config_map map[string]interface{}
+	err2 := json.Unmarshal([]byte(config_json_byteArray), &config_map)
+	fmt.Println(err2)
```
toddgiguere, line 190: see similar comment above, if we should handle this error and "fail and exit"?

tests/pr_test.go
```diff
@@ -163,6 +163,57 @@ func TestRunUpgradeRoksPattern(t *testing.T) {
...
+	new_override_json_string_value, err3 := json.Marshal(override_json_map)
+	new_config_string_value, err4 := json.Marshal(config_map)
+
+	//Compare both the json strings
+	jsonComp, err3 := common.IsJsonEqual(string(new_override_json_string_value), string(new_config_string_value))
+
+	assert.Nil(t, err, "This should not have errored")
+	assert.NotNil(t, output, "Expected some output")
+	assert.NotNil(t, options.LastTestTerraformOutputs, "Expected some Terraform outputs")
+	assert.Nil(t, err1, "Error occured with the override json")
```
toddgiguere, line 208: You should probably handle assertions for errors immediately after they happen, usually when these errors occur you would want to mark the test as failed and then not continue with the test, since you will need the value that errored for the next step. See comments above on handling errors with "require" so that on critical errors the test will fail AND exit if required.

toddgiguere (changes requested):

tests/pr_test.go
Aayush-Abhyarthi: Added

tests/pr_test.go
Aayush-Abhyarthi: made the changes

tests/pr_test.go
Aayush-Abhyarthi: used the require package

tests/pr_test.go
Aayush-Abhyarthi: Changed

tests/pr_test.go
Aayush-Abhyarthi: Made the change

tests/pr_test.go
```diff
@@ -204,6 +204,54 @@ func TestRunUpgradeRoksPattern(t *testing.T) {
...
+	options := setupOptionsRoksPattern(t, "slzovr")
+	options.SkipTestTearDown = true
+	output, err := options.RunTestConsistency()
+	defer options.TestTearDown()
+
+	assert.Nil(t, err, "This should not have errored")
+	assert.NotNil(t, output, "Expected some output")
+	assert.NotNil(t, options.LastTestTerraformOutputs, "Expected some Terraform outputs")
+
+	override_json := filepath.Join(roksPatternTerraformDir, "override.json")
```
toddgiguere, line 219: you won't want to test override if the RunTestConsistency had an error or the outputs is null. You could wrap this entire section in an `if{}` so that you only attempt it if "err" is nil and you have outputs.

toddgiguere (changes requested):

tests/pr_test.go
Aayush-Abhyarthi: added

toddgiguere (approved):

PR #686, opened 2024-01-23
feat: added pre-wired CBR rules

Ak-sky:
### Description

Added support for CBR prewired rules in SLZ.

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


ocofaigh (commented):
@Ak-sky We need more info on what "pre-wired CBR rules" are. Perhaps we need a cloud doc on this @SirSpidey ? And then the variable description for `create_prewired_cbr` could point to it?

Also I don't think we want to run the fscloud CBR module with the default variables values. Thats going to create CBR rules for services that SLZ doesn't even create (such as ICD). We probably need logic to confirm what services SLZ have been set up to deploy, and only create CBR rules for those.
Perhaps worth a deep dive discussion?

Ak-sky:
> @Ak-sky We need more info on what "pre-wired CBR rules" are. Perhaps we need a cloud doc on this @SirSpidey ? And then the variable description for `create_prewired_cbr` could point to it?
> 
> Also I don't think we want to run the fscloud CBR module with the default variables values. Thats going to create CBR rules for services that SLZ doesn't even create (such as ICD). We probably need logic to confirm what services SLZ have been set up to deploy, and only create CBR rules for those. Perhaps worth a deep dive discussion?

I agree @ocofaigh, we had a discussion on this with @vburckhardt initially that if in case CBR rules for other services are required later in future, the flow will already be opened.
  
If needed more discussion on this I can put it in the the deep dive session.

vburckhardt:
It's correct to create global cbr rules in the account including for services not including in landing zone. This is aligned with the design where we lock down all flows in the account - we just need to make sure this is documented. It is just like the AT routing set up for the account. We can use the doc at https://github.com/terraform-ibm-modules/terraform-ibm-cbr/tree/main/modules/fscloud as a base for the content ending up in the public slz doc.

vburckhardt:
perhaps name the variable ```configure_locked_down_cbr_rules_account_wide```

vburckhardt:
Also for the description of the variable - something along those lines: "creates default coarse-grained CBR rules in a given account following a "secure by default" approach - that is: deny all flows by default, except known documented communication in the Financial Services Cloud Reference Architecture"

SirSpidey:
> Also for the description of the variable - something along those lines: "creates default coarse-grained CBR rules in a given account following a "secure by default" approach - that is: deny all flows by default, except known documented communication in the Financial Services Cloud Reference Architecture"

Here's what I'm taking away from the discussion. 

- Update the description of the variable (and perhaps rename it). See suggestion.
- Update the readme list item to something aligned with the variable description.

**Suggestion for description:**
> Creates CBR rules in the account that follow a "secure by default" approach. The rules deny all flows by default except those that are documented in the [VPC reference architecture for IBM Cloud for Financial Services](https://cloud.ibm.com/docs/framework-financial-services?topic=framework-financial-services-vpc-architecture-about).

_If that ref arch link to too specific to VPC, you could link instead to_ [Reference architectures for IBM Cloud for Financial Services](https://cloud.ibm.com/docs/framework-financial-services?topic=framework-financial-services-reference-architecture-overview)

So it doesn't seem that we need

Ak-sky:
I have updated the variable name and [description](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/f0fe0aeba06e8de22def01d07cd28ad24a0a30e1/variables.tf#L1268) as per the suggestion. Should I also add the details from this doc https://github.com/terraform-ibm-modules/terraform-ibm-cbr/tree/main/modules/fscloud to the public slz doc?
@vburckhardt @SirSpidey

ocofaigh (commented):
@Ak-sky what kind of impact is this going to have in our SLZ tests? I think we will end up with conflicting CBR rules in our account when tests run right?

Ak-sky:
> @Ak-sky what kind of impact is this going to have in our SLZ tests? I think we will end up with conflicting CBR rules in our account when tests run right?

Yes, with CBR FSCloud module there will be issues when SLZ test will run parallelly.

ocofaigh:
OK so we need to figure that out then @Ak-sky - perhaps we need to leverage enterprise sub accounts now so every test spins up a new account. We need to solve the apikey issue first - I know there was a feature being added where apikey could be created in the new account for you, and then you could look it up. But I'm assuming its going to be a service ID apikey, which means we couldn't use to provision roks right now. Perhaps add to next deep dive?

Ak-sky:
> parallelly

Yes, I had a discussion on enterprise sub accounts with Todd before and looks like we need this now as we have started consuming CBR FScloud profile and this may cause issues later with other modules consuming CBR FScloud.

I can put this in deep dive and discuss with Todd if there is any POC/investigation I can do to use enterprise sub account.

Ak-sky:
> OK so we need to figure that out then @Ak-sky - perhaps we need to leverage enterprise sub accounts now so every test spins up a new account. We need to solve the apikey issue first - I know there was a feature being added where apikey could be created in the new account for you, and then you could look it up. But I'm assuming its going to be a service ID apikey, which means we couldn't use to provision roks right now. Perhaps add to next deep dive?

Yes, I had a discussion on enterprise sub accounts with Todd before and looks like we need this now as we have started consuming CBR FScloud profile and this may cause issues later with other modules consuming CBR FScloud.

I can put this in deep dive and discuss with Todd if there is any POC/investigation I can do to use enterprise sub account.

Ak-sky:
Also the patterns in test will run in parallel which will be an issue.

ocofaigh:
@Ak-sky Don't run the pipeline until we have a solution as it may impact other tests

PR #687, opened 2024-01-24
feat: add expire_rule support for buckets

akocbek:
### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


ocofaigh (commented):
We probably want to expose this as variable(s) in each of the patterns too, otherwise consumers of our DAs can't use it. cc @toddgiguere for feedback too

toddgiguere (approved):
@akocbek @ocofaigh everything looks good in this PR, but we will need to decide what to do with override.json in the future (I think). I can think of 2 ways to keep override.json updated:
1. as we introduce new features through variable properties, we also add them to the sample override files to showcase
2. override.json sample is derived off of the default landing zone configuration, and would match what is produced from a standard deploy in the `config` output

If option 1 is what we want, then we have done this change correctly.

If option 2 is the direction, then we would want to keep the override.json samples in line with the config produced from a standard landing zone deployment

ocofaigh:
Pipeline failed with `Error: [0m Error: [ERROR] Error in getting bucket info rule: Access Denied` - retrying to see if it happens again..

PR #694, opened 2024-01-31
tests: add new test for override example

jor2:
### Description

Test for override example in landing zone.

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

##### Release notes content

Test the override example in other_tests.go.

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


toddgiguere (changes requested):
I think this test is a duplicate of this one that already exists?
https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/main/tests/pr_test.go#L308

Also this particular test is not really doing anything, I think what it was meant to do is use the `config` output from first  apply as the override for the second apply? But setting that override doesn't take place in this test.

tests/other_test.go
```diff
@@ -48,3 +52,29 @@ func TestRunVSIPatternWithHPCS(t *testing.T) {
...
+
+	if assert.Nil(t, err, "This should not have errored") &&
+		assert.NotNil(t, output, "Expected some output") &&
+		assert.NotNil(t, options.LastTestTerraformOutputs, "Expected some Terraform outputs") {
+
+		// TERRATEST uses its own internal logger.
+		// The "show" command will produce a very large JSON to stdout which is printed by the logger.
+		// We are temporarily turning the terratest logger OFF (discard) while running "show" to prevent large JSON stdout.
+		options.TerraformOptions.Logger = logger.Discard
+		planStruct, planErr := terraform.InitAndPlanAndShowWithStructE(options.Testing, options.TerraformOptions)
```
toddgiguere, line 71: RunTestConsistency() will already run an Apply and then follow it up with a "InitAndPlan" to verify idempotent. I'm not sure what a second "InitAndPlan" will do here?

toddgiguere (changes requested):

jor2:
> I think this test is a duplicate of this one that already exists? https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/main/tests/pr_test.go#L308
> 
> Also this particular test is not really doing anything, I think what it was meant to do is use the `config` output from first apply as the override for the second apply? But setting that override doesn't take place in this test.

That test runs the `quickStartPatternTerraformDir` and does that weird overriding using previous tests output. Should I replace that test with this one?

toddgiguere (approved):
I see the difference now, the other override test I was referring to is more of a "config output" test, and using override. This test is just testing our new example, so I do not think there is an overlap.

PR #704, opened 2024-02-09
fix: update default values inside override.json

akocbek:
### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers

### DIFF
--- a/patterns/mixed/override.json
+++ b/patterns/mixed/override.json
@@ -64,16 +64,6 @@
                         "enable": true,
                         "days": 0,
                         "type": "Glacier"
-                    },
-                    "activity_tracking": {
-                        "read_data_events": true,
-                        "write_data_events": true,
-                        "activity_tracker_crn": "activity-tracker-crn"
-                    },
-                        "metrics_monitoring": {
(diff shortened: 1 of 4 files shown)


ocofaigh (commented):
thanks @akocbek - did you try a test to ensure the override works now with one of the patterns? (you would have to set `override = true` to test it).

akocbek:
we had to remove `activity_tracking` and `metrics_monitoring` because they require valid CRN (e.q. `activity_tracker_crn`) which we don't have it at that stage. If we remove `activity_tracker_crn` or `metrics_monitoring_crn` only then we get the following:
```
│ The given value is not suitable for module.roks_landing_zone.module.landing_zone.var.cos declared at ../../variables.tf:480,1-15: element 0: attribute "buckets": element 0: attribute "metrics_monitoring":
│ attribute "metrics_monitoring_crn" is required.
```

ocofaigh (approved):

Issue #705, opened 2024-02-09
Return the workload and management cluster id as output


vburckhardt:
### Description

Return the management and workload cluster ids as standalone output. This is typically needed higher in the stack, and IBM Cloud Project does not currently allow to create reference on output that are array or map. If there is no cluster named "management" or "workload" (if user changes those via override) leave those output as null.

Aayush-Abhyarthi:
PR merged.

PR #713, opened 2024-02-16
feat: expose cluster ids

Aayush-Abhyarthi:
### Description

[]

### Release required?

- [ ] No release
- [x] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


ocofaigh (commented):
@Aayush-Abhyarthi Can you show proof that outputs worked (screenshot)? And also that they do not output if there is no cluster named "management" or "workload" (if user changes those via override)

ocofaigh:
@Aayush-Abhyarthi Seems your logic is wrong somewhere:
```
│ Error: Invalid index
│ 
│   on ../../outputs.tf line 82, in output "workload_cluster_id":
│   82:   ][0]
│     ├────────────────
│     │ ibm_container_vpc_cluster.cluster is object with no attributes
│     │ var.prefix is "slz-vpc"
│ 
│ The given key does not identify an element in this collection value: the
│ collection has no elements.
╵
╷
│ Error: Invalid index
│ 
│   on ../../outputs.tf line 90, in output "management_cluster_id":
│   90:   ][0]
│     ├────────────────
│     │ ibm_container_vpc_cluster.cluster is object with no attributes
│     │ var.prefix is "slz-vpc"
│ 
│ The given key does not identify an element in this collection value: the
│ collection has no elements.
╵
```

outputs.tf
```diff
@@ -74,6 +74,22 @@ output "cluster_names" {
   ]
 }
 
+output "workload_cluster_id" {
```
toddgiguere, line 77: This is not going to work, since "workload" and "management" are not fixed VPC names. We actually have an input where the consumer can specify any number of VPCs, and their name. It defaults to ["workload","management"] but we expose this at catalog level as well.

The outputs might have to be simply "cluster_ids" and be some type of map or object where the name and ID are output, and you cannot assume how many or what their names are.

toddgiguere (changes requested):

outputs.tf
toddgiguere: For reference, this is the input I was talking about, this is exposed to the user. They can change the names and the number of VPCs, and this is also used when deploying the clusters: https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/main/patterns/roks/variables.tf#L45

Aayush-Abhyarthi: https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/705
The issue specifically asks to output the workload and management cluster ids and NULL if anything other is passed.

toddgiguere: also, if the consumer had used override, the name might not exactly match "-workload-cluster" (or management-cluster). I don't know if this is an issue, might want to check.

outputs.tf
```diff
@@ -74,6 +74,22 @@ output "cluster_names" {
...
+  description = "The id of the workload cluster"
+  value = flatten([
+    for cluster in ibm_container_vpc_cluster.cluster :
+    cluster.id if cluster.name == "${var.prefix}-workload-cluster"
+  ])
+}
+
+output "management_cluster_id" {
+  description = "The id of the management cluster"
```
toddgiguere, line 86: If these are going to be null or empty if workload or management names were not used, perhaps explain all of this in the output descriptions all way through the final output pattern.

Aayush-Abhyarthi: Added in the output

outputs.tf
```diff
@@ -74,6 +74,22 @@ output "cluster_names" {
   ]
 }
 
+output "workload_cluster_id" {
+  description = "The id of the workload cluster. If the cluster name does not exactly match the prefix-workload-cluster pattern it will be empty."
+  value = flatten([
+    for cluster in ibm_container_vpc_cluster.cluster :
+    cluster.id if cluster.name == "${var.prefix}-workload-cluster"
+  ])
```
toddgiguere, line 82: I think the intent here is that the return type of this output is just a single string value with the ID of the workload cluster.

This particular output is (I think) is going to result in a return type of an array containing a single string, which is different. I don't think you want it to return an array with single string. It will also (I think) return an empty array instead of "null" if no matches are found.

I think the issue is the flatten, which you do not need here. Flatten's purpose is to take a bunch of arrays and combine them into single array, which is useful for a `for_each` but not needed when you simply want to return a single string value contained in an array.

toddgiguere (changes requested):

outputs.tf
Aayush-Abhyarthi: made the changes

toddgiguere (approved):

ocofaigh:
Pipeline failed with `[ERROR] Error when creating resource instance: Please contact the Service Provider for this error. [500, Internal Server Error]`
Unrelated to changes in this Pr, so re-running..

PR #716, opened 2024-02-16
feat: added ROKS Quickstart pattern

toddgiguere:
### Description

Addition of new ROKS Quickstart pattern. The purpose of a quickstart pattern is to provide an example of ROKS in a landing zone pattern with minimal resources and public facing endpoints to allow consumers to quickly and cheaply deploy.

This new quickstart will deploy the following resources:
- Two VPCs (management and workload)
- Transit gateway
- All ACLs and SGs are wide open (allow all in and out)
- Public gateways on workload subnets (for cluster outbound)
- 1 workload VPC cluster (4.14_openshift), 2 zones, 1 per zone, flavor bx2.4x16
- Public endpoint enabled for cluster
- 1 KMS with 2 keys for cluster encryption
- 1 COS instance with 1 bucket created by cluster

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


patterns/roks-quickstart/variables.tf
```diff
@@ -0,0 +1,269 @@
...
+}
+
+variable "prefix" {
+  description = "A unique identifier for resources. Must begin with a lowercase letter and end with a lowerccase letter or number. This prefix will be prepended to any resources provisioned by this template. Prefixes must be 16 or fewer characters."
+  type        = string
+  default     = "land-zone-vsi-qs"
+
+  validation {
+    error_message = "Prefix must begin with a lowercase letter and contain only lowercase letters, numbers, and - characters. Prefixes must end with a lowercase letter or number and be 16 or fewer characters."
+    condition     = can(regex("^([a-z]|[a-z][-a-z0-9]*[a-z0-9])$", var.prefix)) && length(var.prefix) <= 16
```
vburckhardt, line 18: The condition in the called module ../roks/module is set to max 13, so the default here does not pass. Also the default should probably not include "vsi"

patterns/roks-quickstart/variables.tf
```diff
@@ -0,0 +1,269 @@
...
+      "resource_group": "",
+      "add_route": false
+   },
+   "clusters": [
+      {
+         "boot_volume_crk_name": "slz-vsi-volume-key",
+         "cos_name": "cos",
+         "kms_config": null,
+         "kube_type": "openshift",
+         "kube_version": "4.14_openshift",
```
vburckhardt, line 51: I think we should put this variable at the top level of the DA, it is going to be a common case for people to change it.
Same for the machine type.

vburckhardt: From an implementation perspective, perhaps we could make the override json a template here. So instead of 
"kube_version": "4.14_openshift",
we would have "kube_version": "${version}" with version being the input variable.

The quick start code would replace the version in the override json and pass it to the ../roks/module.

vburckhardt (commented):
Hi Todd, I put a few comments.

ocofaigh (approved):

ocofaigh:
LGTM - @vburckhardt are you happy with this so we can merge and start working on catalog onboarding?

Issue #645, opened 2023-11-30
Missing kube->kms s2s policy


vburckhardt:
That specific s2s policy does not seem to be created by the module code itself.

On a fresh account:
again.","type":"Authentication"}
 2023/11/30 15:45:51 Terraform apply | 
 2023/11/30 15:45:51 Terraform apply |   with module.roks_landing_zone.module.landing_zone.ibm_container_vpc_cluster.cluster["wxai-management-cluster"],
 2023/11/30 15:45:51 Terraform apply |   on ../../cluster.tf line 34, in resource "ibm_container_vpc_cluster" "cluster":
 2023/11/30 15:45:51 Terraform apply |   34: resource "ibm_container_vpc_cluster" "cluster" {
 2023/11/30 15:45:51 Terraform apply | 
 2023/11/30 15:45:51 Terraform apply | 
 2023/11/30 15:45:51 Terraform apply | Error: Request failed with status code: 401, ServerErrorResponse: {"incidentID":"dc3c5793-4641-44a6-a2d8-699335dbf0f9","code":"E09c0","description":"Not authorized to access the Key Management Service. Create an IBM Cloud IAM authorization policy to give the source Kubernetes Service delegate access to the target Key Management Service, and try again.","type":"Authentication"}
 2023/11/30 15:45:51 Terraform apply | 
 2023/11/30 15:45:51 Terraform apply |   with module.roks_landing_zone.module.landing_zone.ibm_container_vpc_cluster.cluster["wxai-workload-cluster"],
 2023/11/30 15:45:51 Terraform apply |   on ../../cluster.tf line 34, in resource "ibm_container_vpc_cluster" "cluster":
 2023/11/30 15:45:51 Terraform apply |   34: resource "ibm_container_vpc_cluster" "cluster" {
 2023/11/30 15:45:51 Terraform apply | 
 2023/11/30 15:45:51 [1m[31mTerraform APPLY

ocofaigh:
@vburckhardt @Ak-sky This policy is supposed to be auto created when you provision a new cluster. It creates the following global auth policy:
![image](https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/assets/29608224/2f337ff1-800d-45e2-97a9-9273a262eb7b)

So hence that is why we don't create it in our code.

Ak-sky:
[...]

PR #721, opened 2024-02-22
Remove s2s workaround issue comment

Ak-sky:
### Description

Removed the workaround issue description comment as it is mandatory to create s2s kube -> kms auth policy.
Git issue: https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/645

### Release required?

- [x] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


toddgiguere (dismissed):

dynamic_values/config_modules/service_authorizations/service_authorizations.tf
```diff
@@ -59,7 +59,7 @@ module "kms_to_block_storage" {
   ]
 }
 
-# workaround for https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/645
+# Required service authorization access policy for Kubernetes Service and the KMS provider
```
ocofaigh, line 62: Akash I thought we said we would add some extra info in the comment as to why it is required. That basically it only gets auto created if doing cluster data encryption, however it needs to exist before the if doing boot volume encryption, so hence we need to explicitly create it

ocofaigh (changes requested):
see my comment

dynamic_values/config_modules/service_authorizations/service_authorizations.tf
Ak-sky: @ocofaigh I have updated the comment, please check.

ocofaigh (approved):

PR #908, opened 2024-10-23
fix: missing os in override json

Aashiq-J:
### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


patterns/roks/variables.tf
```diff
@@ -289,8 +289,7 @@ variable "cluster_force_delete_storage" {
 
 variable "operating_system" {
   type        = string
-  description = "The operating system of the workers in the default worker pool. If no value is specified, the current default version OS will be used. See https://cloud.ibm.com/docs/openshift?topic=openshift-openshift_versions#openshift_versions_available ."
-  default     = "REDHAT_8_64"
+  description = "The operating system of the workers in the default worker pool. See https://cloud.ibm.com/docs/openshift?topic=openshift-openshift_versions#openshift_versions_available ."
```
ocofaigh, line 292: Isn't the same change needed in `roks/module/variables.tf` ?

ocofaigh (changes requested):
`catalogValidationValues.json.template` will need to be updated too

patterns/roks/variables.tf
Aashiq-J: Thats the tricky part when we directly call the module folder we need to pass a value for the var.operating system even if have a value for it mentioned in the override.json. For example, in quickstart
Screenshot 2024-10-23 at 10 15 55 PM

another option is to remove the var.operating system get the value from json and hardcode "REDHAT_8_64" in the config.tf

@toddgiguere , any idea on a different approach ?

toddgiguere: I think you should leave the default in place for `var.operating_system`, since there is a good default for the default worker pool (which is REDHAT_8_64).

I think the idea of not having a default value was more for the additional worker pools, which this variable is not used for.

ocofaigh (approved):

PR #915, opened 2024-11-12
feat: add support for VNI in vsi module

Aashiq-J:
### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


ocofaigh:
I guess we need to consider how this impact the migration script from v5 -> v6 as user would also have to set the boolean to continue on the legacy, otherwise migration is pointless right?

Aashiq-J:
Yep, we want the new customers to use the new VNI so we can't set `use_legacy_network_interface` to true.
We might want to document in the migration process that they will need to set the `use_legacy_network_interface` to true while upgrading.

Aashiq-J:
Screenshot 2024-11-13 at 9 46 32 AM
Screenshot 2024-11-13 at 9 46 41 AM
Tests which failed in pipeline passes when we set `use_legacy_network_interface ` to true.
We will need to skip the upgrade test since we have `use_legacy_network_interface` as `false` by default.

Issue #863, opened 2024-08-23
Support `user_data` in Console UI


PatrickDuncan:
### Description

> https://cloud.ibm.com/docs/vpc?topic=vpc-user-data
>
> When you create an IBM Cloud® Virtual Servers for Virtual Private Cloud instance, you can specify optional user data that automatically performs common configuration tasks or runs scripts. For more information about user data and usernames, see [IAM access](https://cloud.ibm.com/docs/account?topic=account-userroles).
> 
> VPC uses Cloud-init technology to configure virtual server instances. The User Data field on the New virtual server for VPC page allows users to put in custom configuration options by using cloud-init. Cloud-init supports several formats for configuration data, including yaml in a cloud-config file.

`user_data` is integral VSI instance functionality and we need it to install software on the VSI during startup.

I see that `user_data` is definable using `override_json_string`, but I think its important enough to be a standalone variable.

Users must be able to input something like this in the Deployable Architecture - Configure > Optional panel.
```yaml
#cloud-config
users:
  - name: demouser
    gecos: Demo User
    sudo: ALL=(ALL) NOPASSWD:ALL
    groups: users, admin
    ssh_import_id: None
    lock_passwd: true
```

<img src="https://github.com/user-attachments/assets/d4a92af9-a81f-4245-9b86-0e54040f7769" width=500 />

- [ ] Add `user_data` support as an optional configuration field.
- [ ] The option should be segmented by management and workload VSIs. There is a high chance that the user_data directive will be very different between management and workload VSIs.
- [ ] The help text must link to https://cloud.ibm.com/docs/vpc?topic=vpc-user-data.

### New or affected modules

- https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/tree/main/patterns/vsi
- https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/tree/main/patterns/vsi-quickstart
- https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/tree/main/patterns/vsi-extension

---

By submitting this issue, you agree to follow our [Code of Conduct](https://github.com/terraform-ibm-modules/documentation/blob/main/CODE_OF_CONDUCT.md)

Aashiq-J:
https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/pull/916

Aashiq-J:
PR merged. Closing issue.

PR #916, opened 2024-11-13
feat: add support for the user_data variable

Aashiq-J:
### Description

Issue: https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/issues/863

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [x] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers


variables.tf
```diff
@@ -289,6 +289,20 @@ variable "ssh_keys" {
   }
 }
 
+variable "user_data" {
```
toddgiguere, line 292: We already have a `user_data` field in the `vsi` variable below, so I do not think that we should have a separate variable here.

Is it possible to assign the correct `user_data` to the correct `vsi` object in the pattern instead? That way the override still works and we do not have duplicate variables.

patterns/vsi-quickstart/variables.tf
```diff
@@ -40,6 +40,14 @@ variable "resource_tags" {
   default     = []
 }
 
+variable "user_data" {
```
toddgiguere, line 43: This will not work well in quickstart, as we are hard-coding (in the override) the VSI configurations. If we still want to support user_data in quickstart, we should redesign this pattern a little bit so that user supplied user_data can be inserted into the override json. This does mean a bit of a redesign of this pattern where we would move the override from an input variable with default, and instead have it in the main.tf with variable placeholders. See the `roks-quickstart/main.tf` for an example.

patterns/vsi/variables.tf
```diff
@@ -146,6 +146,14 @@ variable "vsi_per_subnet" {
   default     = 1
 }
 
+variable "user_data" {
```
toddgiguere, line 149: The complex validation that you added at the root level really belongs here at the solution level. If you are going to validate that the map keys here match entries in the `vpcs` list, then you want to do it here at the entry point, and maybe also at the module level of this solution.

Also the description needs to be more detailed explaining to the user what this map is supposed to be (with the key being a VPC name from the `vpcs` variable, and the data being the yaml).

There is something else to mention here: the user_data is usually in yaml format. We need to test that yaml inside an object still works correctly from all deployement types (terraform CLI, schematics, projects) and to make sure the yaml is valid by time it is applied to the VSI.

patterns/vsi/module/main.tf
```diff
@@ -34,6 +34,7 @@ module "landing_zone" {
   teleport_config_data                   = local.env.teleport_config
   teleport_vsi                           = local.env.teleport_vsi
   vpc_placement_groups                   = local.env.vpc_placement_groups
+  user_data                              = var.user_data
```
toddgiguere, line 37: As stated in the root level variables file, I don't think this should be its own separate input variable, as this property already exists within the VSI input map.

Instead, your new input variable for the pattern should be used to set the existing `user_data` property for the individual VSI entries in that map, somewhere in the config.tf: https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/blob/main/patterns/vsi/module/config.tf#L85

dynamic_values.tf
```diff
@@ -33,6 +33,7 @@ module "dynamic_values" {
   skip_kms_kube_s2s_auth_policy          = var.skip_kms_kube_s2s_auth_policy
   skip_all_s2s_auth_policies             = var.skip_all_s2s_auth_policies
   atracker_cos_bucket                    = var.atracker.add_route == true ? var.atracker.collector_bucket_name : null
+  user_data                              = var.user_data
```
toddgiguere, line 36: These changes will not be needed if you assign the user_data to the VSI object out in the pattern level as suggested.

toddgiguere (changes requested):

ocofaigh (changes requested):
Can we update one of the tests to use the `user_data` feature so we at least test it in PR pipeline?

patterns/vsi/variables.tf
toddgiguere, line 149: @ocofaigh do we need to be more specific in this description for how this new map is to be used? The usage would be in the form of:

```
{ a_vpc_name = { user_data = <the data> }, another_vpc_name = { user_data = <the data> }}
```
where the vpc names match what is in the `vpcs` input, typically "workload" and "management". I don't know if this map format is obvious to end user?

ocofaigh: The description is pointing to a cloud doc which explains more (its not yet published)

PR #949, opened 2025-02-08
fix: force vpc flow logs to wait for auth policy sleep

toddgiguere:
### Description

Added a reference to the `time_sleep` resource of bucket auth polices to the VPC Module input for flow log bucket, to implicitly force the flow logs to wait for all bucket auth policies to complete first.

This fixes a random timing issue in which flow logs are getting set up slightly before the auth policy for buckets are finished.

### Release required?

- [ ] No release
- [x] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers

### DIFF
--- a/atracker.tf
+++ b/atracker.tf
@@ -31,7 +31,7 @@ resource "ibm_atracker_target" "atracker_target" {
   target_type = "cloud_object_storage"
 
   # Wait for buckets and auth policies to ensure successful provision
-  depends_on = [ibm_cos_bucket.buckets, ibm_iam_authorization_policy.policy, ibm_iam_authorization_policy.cos_bucket_policy]
+  depends_on = [ibm_cos_bucket.buckets, ibm_iam_authorization_policy.policy, time_sleep.wait_for_authorization_policy_buckets]
 }
 
 resource "ibm_atracker_route" "atracker_route" {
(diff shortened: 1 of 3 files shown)


toddgiguere:
I believe this is working, there was only one error on the last run and it was a failure in schematics service.

Will keep trying, but I think this fix will resolve the timing issue on auth policies.

toddgiguere:
@ocofaigh in the last test run (which is the current one right now) the only test to fail was `TestRunUpgradeVpcPattern`, which failed on `main` branch flow logs due to the timing issue I'm trying to fix. The problem with the upgrade tests here is that the main branch will still have the random occurance of the timing issue.

I'm leaving this latest run here, I think we can safely skip the upgrade tests again? In previous runs here on this PR this latest failure did not happen, so I think we will be at the mercy of the random fails (timing issue) on all the upgrade tests.

ocofaigh:
Can we merge https://github.com/terraform-ibm-modules/terraform-ibm-landing-zone/pull/948 first so that the validation done uses schematics?

toddgiguere:
Only errors in latest run were upgrade errors, and they were all the same:
```
Resource(s) identified to be destroyed 
Name: wait_for_authorization_policy_buckets
Address: module.vpc_landing_zone.module.landing_zone.time_sleep.wait_for_authorization_policy_buckets
Actions: [delete]
```
This was the only resource reported by the upgrade tests, and it is expected, as I've changed the `wait_for_authorization_policy_buckets` from a single instance into a `for_each` multiple resource, one for each bucket.

I'm going to skip upgrade tests since these are expected, and to avoid the `main` branch failing for the timing reason we are fixing here.

PR #955, opened 2025-02-19
chore: add the scripts back due to a bug in projects

Aashiq-J:
### Description

There is an ongoing bug in the projects which picks runs the migration script even if we removed it in the newer version which is causing an error. So, for now adding the script back till the issue is fixed.

### Release required?

- [ ] No release
- [ ] Patch release (`x.x.X`)
- [ ] Minor release (`x.X.x`)
- [ ] Major release (`X.x.x`)

### Run the pipeline

```
/run pipeline
```

### Checklist for reviewers

### For mergers

### DIFF
--- a/patterns/roks-quickstart/scripts/migration-pre-ansible-playbook.yaml
+++ b/patterns/roks-quickstart/scripts/migration-pre-ansible-playbook.yaml
@@ -1,0 +1,103 @@
+- name: migration pre playbook
+  hosts: localhost
+  tasks:
+    - name: get running ansible env variables
+      set_fact:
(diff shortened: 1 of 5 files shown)


ocofaigh:
I'll hold off on this - I would prefer get the fix from Project first. But use this as backup if we need to force a new release before then

ocofaigh (changes requested):
@Aashiq-J We don't have to add the exact migration script back in. As a workaround we can just have a script that does nothing. Want to update this?

ocofaigh (approved):

PR #1045, opened 2025-09-03
Update deploy-arch-ibm-slz-vpc-standard.md

pranjali-mundra:
### Description
Update teh references to the DA name and the variation name to Cloud Foundation for VPC and Standard FS edition.

### DIFF
--- a/reference-architectures/deploy-arch-ibm-slz-vpc-standard.md
+++ b/reference-architectures/deploy-arch-ibm-slz-vpc-standard.md
@@ -1,10 +1,10 @@
 ---
 
 copyright:
-  years: 2023, 2024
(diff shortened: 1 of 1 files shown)


reference-architectures/deploy-arch-ibm-slz-vpc-standard.md
```diff
@@ -16,7 +16,7 @@ version: 6.6.0
 
 # Whether the reference architecture is published to Cloud Docs production.
 # When set to false, the file is available only in staging. Default is false.
-production: true
+production: false
```
ocofaigh, line 19: why is this false? It is in production

